PR #13, opened 2022-10-01
:sparkles: fix loop

maxlerebourg:


mathieuHa (changes requested):
Traefik logs files are usually written in /var/log/traefik/access.log and not traefik.log

PR #15, opened 2022-10-11
:bento: remove alone mode

maxlerebourg:


bouncer.go
```diff
@@ -12,53 +12,43 @@ import (
...
+	CrowdsecLapiKey        string `json:"crowdsecLapiKey,omitempty"`
+	UpdateIntervalSeconds  int64  `json:"updateIntervalSeconds,omitempty"`
+	DefaultDecisionSeconds int64  `json:"defaultDecisionSeconds,omitempty"`
 }
 
 // CreateConfig creates the default plugin configuration.
 func CreateConfig() *Config {
 	return &Config{
 		Enabled:                false,
-		CrowdsecMode:           streamMode,
```
mathieuHa, line 55: Why there we change the default mode, please change it in a separate PR and explain why ?

mathieuHa (changes requested):

mathieuHa:
Changes seems OK, will tests some use case before we can merge this

PR #24, opened 2022-10-16
Create Environnement to work on redis cache

mathieuHa:


mathieuHa:
First version with redis support

Used the most popular library which in fact is not compatible: (unsafe dependancy)

will try with another: https://github.com/gomodule/redigo

PR #71, opened 2022-12-31
:bug: fix all incoherence

maxlerebourg:


pkg/configuration/configuration.go
```diff
@@ -144,37 +145,24 @@ func ValidateParams(config *Config) error {
 	}
 	// We need to either have crowdsecLapiKey defined or the BouncerCert and Bouncerkey
 	if lapiKey == "" && (certBouncer == "" || certBouncerKey == "") {
-		return fmt.Errorf("CrowdsecLapiKey || (CrowdsecLapiTLSCertificateBouncer && CrowdsecLapiTLSCertificateBouncerKey): cannot be both empty")
```
mathieuHa, line 147: Since values are string, it makes more sense to call empty rather than false.

pkg/configuration/configuration.go
```diff
@@ -144,37 +145,24 @@ func ValidateParams(config *Config) error {
...
+		return fmt.Errorf("CrowdsecLapiKey || (CrowdsecLapiTLSCertificateBouncer && CrowdsecLapiTLSCertificateBouncerKey): cannot be both false")
 	} else if lapiKey != "" && (certBouncer == "" || certBouncerKey == "") {
-		// check LAPIKey
-		lapiKey = strings.TrimSuffix(lapiKey, "\n")
-		err = validateParamsAPIKey(lapiKey)
-		if err != nil {
+		lapiKey = strings.TrimSpace(lapiKey)
```
mathieuHa, line 150: Why this change ?  
This does not solve the same issue.  
We could consider removing spaces + end of line caracters

pkg/configuration/configuration.go
```diff
@@ -183,89 +171,14 @@ func validateParamsAPIKey(lapiKey string) error {
 //	tchar = "!" / "#" / "$" / "%" / "&" / "'" / "*" / "+" / "-" / "." /
 //	        "^" / "_" / "`" / "|" / "~" / DIGIT / ALPHA
 //	token = 1*tchar
-func validHeaderFieldByte(b byte) bool {
```
mathieuHa, line 186: This function was in the standard go library of net.  
I guess this form was more perfomant somehow

pkg/ip/ip.go
```diff
@@ -23,9 +23,8 @@ type Checker struct {
 func NewChecker(trustedIPs []string) (*Checker, error) {
 	checker := &Checker{}
 
-	for _, ipMask := range trustedIPs {
-		// remove leading and trailing spaces
```
mathieuHa, line 27: Could you add the comment removed ?

pkg/configuration/configuration_test.go
```diff
@@ -185,6 +192,30 @@ func Test_validateParamsRequired(t *testing.T) {
...
+	tests := []struct {
+		name    string
+		args    args
+		wantErr bool
+	}{
+		{name: "Validate minimal config", args: args{lapiKey: "test!#$%&'*+-.^_`|~"}, wantErr: false},
+		{name: "Not validate a bad crowdsec scheme", args: args{lapiKey: "test@"}, wantErr: true},
+		{name: "Not validate a bad crowdsec mode", args: args{lapiKey: "test("}, wantErr: true},
+		{name: "Not validate a bad update interval seconds", args: args{lapiKey: "test["}, wantErr: true},
+		{name: "Not validate a bad default decision seconds", args: args{lapiKey: "test?"}, wantErr: true},
```
mathieuHa, line 208: Name is incoherent with the test

PR #82, opened 2023-01-30
81 bug stream mode stops blocking

maxlerebourg:


mathieuHa (commented):
Are you sure that it will relaunch ?

HandleStream is not longer running in it's go routine

PR #84, opened 2023-02-27
:sparkles: handle isHealthy in the main function and log error became…

maxlerebourg:
… debug

### DIFF
--- a/bouncer.go
+++ b/bouncer.go
@@ -143,6 +143,7 @@ func New(ctx context.Context, next http.Handler, config *configuration.Config, n
 	}
 	bouncer.cacheClient.New(config.RedisCacheEnabled, config.RedisCacheHost)
 
+	//nolint:nestif
 	if (config.CrowdsecMode == configuration.StreamMode || config.CrowdsecMode == configuration.AloneMode) && ticker == nil {
 		if config.CrowdsecMode == configuration.AloneMode {
 			err = getToken(bouncer)
--- a/bouncer.go
+++ b/bouncer.go
@@ -151,11 +151,18 @@ func New(ctx context.Context, next http.Handler, config *configuration.Config, n
 				return nil, err
 			}
 		}
-		ticker = startTicker(config, func() {
-			handleStreamCache(bouncer)
-		})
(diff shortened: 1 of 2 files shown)


bouncer.go
```diff
@@ -218,7 +226,7 @@ func (bouncer *Bouncer) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 		if isCrowdsecStreamHealthy {
 			bouncer.next.ServeHTTP(rw, req)
 		} else {
-			logger.Error(fmt.Sprintf("ServeHTTP isCrowdsecStreamHealthy:false ip:%s", remoteIP))
+			logger.Debug(fmt.Sprintf("ServeHTTP isCrowdsecStreamHealthy:false ip:%s", remoteIP))
```
mathieuHa, line 229: Do we still get any error logs when we're not in Debug mode about stream not behing healthy ?

I think we should find a way to still get that information so we can know something is wrong without adding debug mode.

But it should only be printed once for instance.

bouncer.go
```diff
@@ -151,11 +152,18 @@ func New(ctx context.Context, next http.Handler, config *configuration.Config, n
...
 			}
 		}
+		if err := handleStreamCache(bouncer); err != nil {
+			return nil, err
+		}
+		isStartup = false
 		ticker = startTicker(config, func() {
-			handleStreamCache(bouncer)
+			if err := handleStreamCache(bouncer); err != nil {
+				isCrowdsecStreamHealthy = false
```
mathieuHa, line 161: Here, the log has been moved from the inner function to the caller, which is good because errors can be handled this way better.

But I assume it will generate as much more logs than before.

mathieuHa: I would suggest that we add a parameter that register the number of time that the stream has been `unhealthy`

This will be a parameter of a function calculating the refresh period.
Let say period = (initial_period + 60 sec * nb_failed_healthy)

Once stream is healthy this parameter would be resetted to 0.

This would reduce the number of logs and the pressure until the stream is healthy again

mathieuHa (changes requested):

PR #89, opened 2023-03-05
🐛Bump simple redis to 1.0.6 to fix bug hang with password, update doc…

mathieuHa:
… on redis


mathieuHa:
seems fix, but exemple does weird behavior.

Looks like only first redis container is contacted by each of services

@maxlerebourg

PR #110, opened 2023-08-18
:wrench: Add conf variable: CrowdsecStreamTimeout

mathieuHa:


mathieuHa:
We should add a test to check if the value can be converted as int

mathieuHa:
other solution than fmt.sscan would be to import package `strconv`

PR #112, opened 2023-09-01
🐛 env values need to be string

JosephKav:
❯ helm -n crowdsec upgrade --install crowdsec crowdsec/crowdsec --values=/opt/helm/values.yml
> Error: UPGRADE FAILED: failed to create resource: Deployment in version "v1" cannot be handled as a Deployment: json: cannot unmarshal bool into Go struct field EnvVar.spec.template.spec.containers.env.value of type string


mathieuHa:
Alright, have'nt tested in a bit with the file in the exemple, thanks for the correction.  

I'll merge this now.

PR #141, opened 2024-04-01
🐛 fix bug with recaptcha

maxlerebourg:


mathieuHa (approved):
Tested, and added some doc for recpatcha

PR #145, opened 2024-04-03
130 feature custom html error page integration on 403 like captcha but for ban

mathieuHa:


ban.html
```diff
@@ -326,4 +326,5 @@ <h1 class="text-2xl lg:text-3xl xl:text-4xl">CrowdSec Access Forbidden</h1>
     </div>
   </div>
 </body>
-</html>
+
+</html>
```
maxlerebourg, line 330: nop

pkg/configuration/configuration.go
```diff
@@ -67,8 +67,8 @@ type Config struct {
 	RedisCachePassword                       string   `json:"redisCachePassword,omitempty"`
 	RedisCachePasswordFile                   string   `json:"redisCachePasswordFile,omitempty"`
 	RedisCacheDatabase                       string   `json:"redisCacheDatabase,omitempty"`
-	BanHTMLFilePath                          string   `json:"banHtmlFilePath,omitempty"`
 	CaptchaHTMLFilePath                      string   `json:"captchaHtmlFilePath,omitempty"`
+	BanHTMLFilePath                          string   `json:"banHtmlFilePath,omitempty"`
```
maxlerebourg, line 71: Why this change ?

maxlerebourg (approved):

pkg/configuration/configuration.go
mathieuHa: Idea was to not use template for static

PR #153, opened 2024-04-18
✨ Add grace period to reach LAPI without blocking further queries

mathieuHa:


bouncer.go
```diff
@@ -279,8 +283,12 @@ func (bouncer *Bouncer) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 
 	// Right here if we cannot join the stream we forbid the request to go on.
 	if bouncer.crowdsecMode == configuration.StreamMode || bouncer.crowdsecMode == configuration.AloneMode {
+		//nolint:gocritic
 		if isCrowdsecStreamHealthy {
 			handleNextServeHTTP(bouncer, remoteIP, rw, req)
+		} else if bouncer.nbFailedStreamUpdate < bouncer.maxFailedStreamUpdate || bouncer.maxFailedStreamUpdate == -1 {
+			bouncer.nbFailedStreamUpdate++
```
maxlerebourg, line 290: If the `maxFailedStreamUpdate` is set to 2 and 3 people try to access a route, you block the third and all the next ones

bouncer.go
```diff
@@ -517,6 +525,7 @@ func handleStreamCache(bouncer *Bouncer) error {
 	}
 	bouncer.log.Debug("handleStreamCache:updated")
 	isCrowdsecStreamHealthy = true
+	bouncer.nbFailedStreamUpdate = 0
```
maxlerebourg, line 528: do with a global var like `isCrowdsecStreamHealthy` or replace `isCrowdsecStreamHealthy` by `bouncer.isCrowdsecStreamHealthy`

bouncer.go
mathieuHa: That's true, we need to increment only once every stream update

bouncer.go
mathieuHa: if there was no reason to use global var in the first place, i suggest add an attribute for the bouncer

bouncer.go
```diff
@@ -360,14 +360,13 @@ func handleNextServeHTTP(bouncer *Bouncer, remoteIP string, rw http.ResponseWrit
...
 
 func handleStreamTicker(bouncer *Bouncer) {
-	bouncer.log.Debug(fmt.Sprintf("handleStreamTicker:start isCrowdsecStreamHealthy:%t", isCrowdsecStreamHealthy))
 	if err := handleStreamCache(bouncer); err != nil {
-		bouncer.log.Debug(fmt.Sprintf("handleStreamTicker:%d failure to reach crowdsec", updateFailure))
-		if updateFailure >= bouncer.updateMaxFailure && bouncer.updateMaxFailure != -1 {
+		bouncer.log.Debug(fmt.Sprintf("handleStreamTicker updateFailure:%d isCrowdsecStreamHealthy:%t %s", updateFailure, isCrowdsecStreamHealthy, err.Error()))
+		if bouncer.updateMaxFailure != -1 && updateFailure >= bouncer.updateMaxFailure && isCrowdsecStreamHealthy {
 			isCrowdsecStreamHealthy = false
+			bouncer.log.Error(fmt.Sprintf("handleStreamTicker:error updateFailure:%d %s", updateFailure, err.Error()))
```
mathieuHa, line 367: why not use the same format of logs as just below ?

maxlerebourg (approved):

PR #189, opened 2024-09-24
✨ add remediation header when plugin made decision

maxlerebourg:


mathieuHa (approved):
Okay this looks good, only thing that could be added maybe is a test with the header set and without it

PR #192, opened 2024-10-01
✨ Add Traefik Crowdsec Plugin Basic User Agent

mathieuHa:
Following the [rfc](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent) the idea would be to have something like:

"User-Agent": "Traefik Crowdsec Plugin / <plugin-version>"

We'll look if you can get this information from Traefik context


maxlerebourg (approved):

grasshide:
Not sure if it is only me, but my log fills up with messages like this:
{
  "time": "2024-10-20T16:03:01Z",
  "level": "warning",
  "msg": "bad user agent 'Crowdsec bouncer Traefik Plugin'",
  "ip": "192.168.240.16",
  "name": "traefik-bouncer"
}

PR #208, opened 2025-01-24
✨ [BREAKING-CHANGE] Add CrowdsecAppsecBodyLimit

maxlerebourg:


pkg/configuration/configuration.go
```diff
@@ -328,6 +330,9 @@ func validateParamsRequired(config *Config) error {
 	if config.UpdateMaxFailure < -1 {
 		return errors.New("UpdateMaxFailure: cannot be less than -1")
 	}
+	if config.CrowdsecAppsecBodyLimit < 0 {
```
mathieuHa, line 333: less or equal

mathieuHa: Do we want to allow not to process the body ?

maxlerebourg: Exactly, I put `if bouncer.appsecBodyLimit > 0 && httpReq.Body != nil && httpReq.ContentLength > 0 {` to prepare post body to appsec, we can disable it with 0

mathieuHa: Alright that's good

mathieuHa (approved):

PR #217, opened 2025-02-15
✨ feat(logs) add supports write logs to files

mathieuHa:


pkg/configuration/configuration.go
```diff
@@ -98,7 +102,8 @@ func contains(source []string, target string) bool {
 func New() *Config {
 	return &Config{
 		Enabled:                        false,
-		LogLevel:                       "INFO",
+		LogLevel:                       LogERROR,
```
maxlerebourg, line 105: You change the default value ?

pkg/configuration/configuration.go
```diff
@@ -262,6 +267,16 @@ func ValidateParams(config *Config) error {
 		}
 	}
 
+	// Check logging configuration
+	if !(config.LogLevel == LogERROR || config.LogLevel == LogINFO || config.LogLevel == LogDEBUG) {
```
maxlerebourg, line 271: if !contains([]string{LogERROR, LogINFO, LogDEBUG}, config.CrowdsecMode) {
  return errors.New("LogLevel should be one of 'INFO', 'ERROR', 'DEBUG'")
}

pkg/logger/logger.go
```diff
@@ -16,15 +18,29 @@ type Log struct {
...
+			_ = fmt.Errorf("LogFilePath is not writable %w", err)
+		} else {
+			logInfo.SetOutput(logFile)
+			logError.SetOutput(logFile)
+			if logLevel == "DEBUG" {
+				logDebug.SetOutput(logFile)
+			}
+		}
+	}
+
```
maxlerebourg, line 43: Add a log to explain why we do setOutput twice

PR #220, opened 2025-02-18
Minor fixes to captcha

david-garcia-garcia:
A couple of nitpicks I found while setting up captcha support:

*  Add error to logs when remediation is set to captcha, but captcha client is not valid.
* Fix styles for default captcha html so that the actual captcha is centered in screen

This is what it looked before the fix:

image

I found it weird that the docs have actual images of hcaptcha setup instead of google recaptcha, yet they state hcaptcha is not supported.

### DIFF
--- a/captcha.html
+++ b/captcha.html
@@ -276,6 +276,13 @@
         line-height: 2.5rem
       }
     }
+
+    div.g-recaptcha {
+      margin: 0 auto;
+      width: 304px;
+      margin-top: 20px;
+      margin-bottom: 20px;
+    }
   </style>
   <script src="{{ .FrontendJS }}" async defer></script>
 </head>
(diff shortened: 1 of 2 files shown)


captcha.html
```diff
@@ -276,6 +276,13 @@
         line-height: 2.5rem
       }
     }
+
+    div.g-recaptcha {
+      margin: 0 auto;
+      width: 304px;
+      margin-top: 20px;
+      margin-bottom: 20px;
```
maxlerebourg, line 284: Your margins are very weird, maybe you can use `margin: 20px auto 20px`

Additionnaly, not very sure about set a width to this div.

You can play with `display: flex; justify-items: center;`

maxlerebourg:
I've corrected this CSS problem in another branch, for the log, I don't see the point if the user hasn't consciously enabled the captcha.
I will close this PR in favor of https://github.com/maxlerebourg/crowdsec-bouncer-traefik-plugin/pull/225
I saw you opened another PR to report a blocked user to crowdsec LAPI, I'll be happy to review it when it's ready.

PR #226, opened 2025-03-24
feat: support altcha as a captcha provider

holysoles:
This PR adds [ALTCHA](https://altcha.org/) as a captcha provider. ALTCHA is a privacy friendly, FOSS solution for captcha that validates users based on a proof-of-work (PoW) challenge done by the client.

## Notes:

- ALTCHA does not follow the same method as other providers as replacing a form div with their widget. This is solved in the PR by conditionally including an additional JS snippet that sets up the altcha widget **if** altcha is the chosen provider.
- An API key is for **all** API calls, which means that creating a challenge needs an API key. I've implemented retrieval of the challenge data server side in an attempt to protect the API key. This is then passed into the template and used to configure the widget.
- The official API **doesn't** support the existing POST Form request send to verify the challenge. This means we need separate logic to send a JSON body with a POST request to validate the challenge result.
- I've exposed the Challenge and Validation URLs as configuration parameters. Since ALTCHA is fully open source, it is possible to host your own server that meets their API spec.

## Testing:

I am happy to provide further assistance in others testing or demonstrating testing evidence myself, but the following should work:

- Generate a free API key at [this page](https://altcha.org/docs/api/free_api_keys/) (it takes awhile, be patient).
  - grab the `apiKey` value that is returned, we'll need that
- run the `examples/captcha/docker-compose.yml` example with the following labels set on `whoami`:
```yaml
      - "traefik.http.middlewares.crowdsec.plugin.bouncer.captchaProvider=altcha"
      - "traefik.http.middlewares.crowdsec.plugin.bouncer.captchaReferer=<domain entered on ALTCHA form>"
      - "traefik.http.middlewares.crowdsec.plugin.bouncer.captchaSecretKey=<apiKey value>"
```
- Start the containers: `make run_captcha`
- Set a captcha decision for your IP: `docker exec crowdsec cscli decisions add --ip 10.0.0.10 -d 10m -t captcha` (my IP was 172.18.0.1)
- Test that ALTCHA is used as the captcha provider for `http://localhost:8000/foo`


mathieuHa:
Hi @holysoles,
Thanks for the PR, we will review it in the following days

pkg/captcha/captcha.go
```diff
@@ -115,21 +156,50 @@ type responseProvider struct {
...
+}
+
 // Validate Verify the captcha from provider API.
 func (c *Client) Validate(r *http.Request) (bool, error) {
 	if r.Method != http.MethodPost {
 		c.log.Debug("captcha:Validate invalid method: " + r.Method)
 		return false, nil
 	}
-	var response = r.FormValue(captcha[c.provider].key + "-response")
+	var response = r.FormValue(captcha[c.provider].key)
```
maxlerebourg, line 173: Have you test this change on other providers ?

pkg/captcha/captcha.go
```diff
@@ -142,11 +212,48 @@ func (c *Client) Validate(r *http.Request) (bool, error) {
...
+	if err != nil {
+		return err
+	}
+	defer res.Body.Close()
+	err = json.NewDecoder(res.Body).Decode(&cd)
+	if res.StatusCode != 200 {
+		return errors.New("error retrieving challenge data: (" + res.Status + ") " + cd.Error)
+	}
+	return err
+}
```
maxlerebourg, line 259: Please, fix lint in your branch

pkg/captcha/captcha.go
holysoles: previously i was only testing i could load the other widgets since I didn't think i modified the validation behavior.

It has now been tested with cloudflare turnstile, and looks good on the latest commit

pkg/captcha/captcha.go
holysoles: should be fixed, ran `make` locally

holysoles:
comments addressed in latest commit, let me know if you'd like me to squash

mathieuHa:
Hi @holysoles, the PR looks good, and we could finally support a captcha solution that can be self hosted and not in the cloud.  

On my end it looks good, I'll have to do a bit of testing before we can move forward.
I also have to convince @maxlerebourg, this adds quite some complexity compared to other 3 standardized providers.

If anyone is interested for this addition please give us a thumb up here

mathieuHa:
Hi @holysoles,

I've started to test the provider:
I created the API key using your link:

```json
{
  "hostname": "localhost:8000",
  "apiKey": "ckey_XXXXXXXXXXX",
  "secret": "csec_XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX,
 ....
}
```
I should use apiKey or secret for `captchaSecretKey`.
Also, `captchaSiteKey` should be empty ?

I used:
```
 - "traefik.http.middlewares.crowdsec.plugin.bouncer.captchaProvider=altcha"
 - "traefik.http.middlewares.crowdsec.plugin.bouncer.captchaReferer=localhost:8000"
 - "traefik.http.middlewares.crowdsec.plugin.bouncer.captchaSecretKey=ckey_XXXXXXXXXXX"
```

By default is this the minimal configuration ?
I don't see how it can work if the following variables are not set:
```
- `captchaChallengeURL`
- `captchaValidationURL`
```
Also if we want to use the cloud hosted version, I see there is a EU and a US hosted one. How do we fill this ?

I think in the exemple, we need to update the `captcha.html` file with the one on the root dir of the project.

holysoles:
@mathieuHa  thanks for testing!

> I should use apiKey or secret for captchaSecretKey.

You should use the `apiKey` as `captchaSecretKey` in the configuration, the `secret` provided by captcha is only needed  when validating a client challenge (which the API does in this case)

>  Also, captchaSiteKey should be empty ?

Yes, it is not needed for this provider

> By default is this the minimal configuration ?
> I don't see how it can work if the following variables are not set:

Yes this is the default configuration, if someone wants to use the EU instance of altcha's API. If configuration values for `captchaChallengeURL` and `captchaValidationURL` are not provided, defaults are provided in `captcha.go` [at L69 and L70](https://github.com/maxlerebourg/crowdsec-bouncer-traefik-plugin/blob/ba23cab551ebe840ceb6ccde8543cba378293c00/pkg/captcha/captcha.go#L69)

> Also if we want to use the cloud hosted version, I see there is a EU and a US hosted one. How do we fill this ?

It is set to use EU by default, I figured that would be preferred as a default given the better privacy laws. This is an example configuration for using the US one if desired:

```yaml
      - "traefik.http.middlewares.crowdsec.plugin.bouncer.captchaProvider=altcha"
      - "traefik.http.middlewares.crowdsec.plugin.bouncer.captchaReferer=<domain entered on ALTCHA form>"
      - "traefik.http.middlewares.crowdsec.plugin.bouncer.captchaSecretKey=<apiKey value>"
      - "traefik.http.middlewares.crowdsec.plugin.bouncer.captchaChallengeURL=https://us.altcha.org/api/v1/challenge"
      - "traefik.http.middlewares.crowdsec.plugin.bouncer.captchaValidationURL=https://us.altcha.org/api/v1/challenge/verify"
```

> I think in the example, we need to update the captcha.html file with the one on the root dir of the project.

Yes you are correct, i was using the one in the project root in my testing. I will push a commit that fixes this

> I think this line is a bit strange: captcha:Validate invalid method: GET

This is definitely odd. I will try to investigate

holysoles:
The captcha example HTML file has been updated.

Regarding the log message `captcha:Validate invalid method: GET`: I'm pretty sure its because `captcha.Client.ServeHTTP()` [tries to validate any request is receives](https://github.com/maxlerebourg/crowdsec-bouncer-traefik-plugin/blob/main/pkg/captcha/captcha.go#L79), but `Validate()` wants only POST requests.

It doesn't seem like I introduced this behavior, I can reproduce it on `main`. However I did address this in my most recent commit. Let me know if you have any feedback 😄, this should be ready for another look whenever you have time.

maxlerebourg:
It is now possible to use a custom captcha. If you wish to add Altcha, you can create a new example with a new captcha.html file. Thank you very much for your contribution. I used your code as inspiration for a more general use case.

PR #236, opened 2025-04-28
Fix query of the GET /decisions endpoint

lluisd:
banned=true query parameter doesn't exist, use type=ban based on LAPI dcoumentation https://crowdsecurity.github.io/api_doc/lapi/#/Remediation%20component/getDecisions

testing the endpoint 
image


maxlerebourg:
You spot a good one, but if we merge this, the captcha logic will break, please update your PR to remove the `banned=true` part.

lluisd:
Already removed

maxlerebourg (approved):

PR #250, opened 2025-06-30
✨ Add env for RemediationStatusCode

maxlerebourg:


pkg/configuration/configuration.go
```diff
@@ -119,6 +120,7 @@ func New() *Config {
 		UpdateIntervalSeconds:          60,
 		UpdateMaxFailure:               0,
 		DefaultDecisionSeconds:         60,
+		RemediationStatusCode:          http.StatusForbidden,
```
mathieuHa, line 123: We could add validation of 100<RemediationStatusCode<500

PR #296, opened 2025-12-08
Add support for injecting request header value into ban HTML template

highpingblorg:
## Overview

Adds a parameter `displayRequestHeader` where you can specify a request header to whose value can be injected into the ban HTML template via `.CustomHeader`.

## Why

image

A use-case would be to emulate the feature like with Cloudflare where a user who is blocked has a `Cloudflare Ray ID`.

How this would work with Traefik+Crowdsec is by having Traefik appending a request header like `X-Trace-ID` or `X-Request-ID` (for example: https://github.com/trinnylondon/traefik-add-trace-id). This would allow users to contact website administrators, in case they believe they have been wrongfully blocked, using the provided ID that would help administrators easily track it down.

### DIFF
--- a/README.md
+++ b/README.md
@@ -497,7 +497,7 @@ make run
   - string
   - default: ""
   - Path where the ban html file is stored (default empty ""=disabled)
-- TraceCustomHeader
+- TraceHeadersCustomName
   - string
   - default: ""
   - Request Header name whose value to inject in ban HTML response (default empty ""=disabled)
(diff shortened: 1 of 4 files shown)


maxlerebourg:
I like the idea but the naming is not coherent with the project:  
Can you use `TraceHeaderCustomName` instead of `DisplayRequestHeader` ?  
And be careful the lint won't pass.

maxlerebourg:
Look at others param names: 
- `RemediationHeadersCustomName`
- `ForwardedHeadersCustomName`
Please use `TraceHeadersCustomName`

PR #300, opened 2025-12-11
🐛 fix start up config error for appsec

### LINKED ISSUE #298: [BUG] release 1.4.7 is currently broken
**Describe the bug** 🐛
After updating from 1.4.6 to 1.4.7 appsec no longer worked. I had to add both crowdsecAppsecScheme as well as crowdsecAppSecKey to my traefik config. The latter is not documented.

**Expected behavior** 👀
Document crowdsecAppSecKey in README.md configuration section and document both changes in Variables section of the README.md.
It can be considered to default to http for crowdsecAppsecScheme, just like crowdsecLapiScheme. It can also be considered to use CrowdsecLapiKey for crowdsecAppSecKey, when the latter is empty. This will help to not break existing installs

maxlerebourg:


bouncer.go
```diff
@@ -173,6 +179,9 @@ func New(_ context.Context, next http.Handler, config *configuration.Config, nam
 			return nil, errKey
 		}
 		config.CrowdsecLapiKey = apiKey
+		if config.CrowdsecAppsecKey == "" {
```
mathieuHa: we use the same key for lapi and appsec ?

mathieuHa: it's not documented, so it will always be empty here

mathieuHa (approved):
LGTM

PR #305, opened 2025-12-30
Do not validate Crowdsec LAPI authentication credentials if bouncer is in Appsec mode

highpingblorg:
Using `CrowdsecAppsecKey` or `CrowdsecAppSecKeyFile` without `CrowdsecLapiKey` causes the bouncer to error out and throw `CrowdsecLapiKey || (CrowdsecLapiTLSCertificateBouncer && CrowdsecLapiTLSCertificateBouncerKey): cannot be all empty`.

With the introduction of the `CrowdsecAppsecKey` and `CrowdsecAppsecKeyFile`, it would make the most visual and logical sense for these parameters to be specified when the bouncer is set to `appsec` mode. Otherwise you would be specifying a LAPI parameter for AppSec when you aren't using the LAPI in the middleware at all.

Just adding a condition to not throw the error relating to LAPI authentication if the mode is set to `appsec` does the trick.

### DIFF
--- a/pkg/configuration/configuration_test.go
+++ b/pkg/configuration/configuration_test.go
@@ -205,23 +205,24 @@ func Test_validateParamsRequired(t *testing.T) {
 
 func Test_validateParamsAPIKey(t *testing.T) {
 	type args struct {
-		lapiKey string
+		lapiKey   string
+		paramName string
 	}
 	tests := []struct {
 		name    string
 		args    args
 		wantErr bool
 	}{
-		{name: "Validate all the valid characters", args: args{lapiKey: "test!#$%&'*+-.^_`|~"}, wantErr: false},
-		{name: "Not validate a @", args: args{lapiKey: "test@"}, wantErr: true},
-		{name: "Not validate a (", args: args{lapiKey: "test("}, wantErr: true},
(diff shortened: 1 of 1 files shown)


maxlerebourg (dismissed):

mathieuHa:
Hi @highpingblorg,

This make sense.  
I think it would be nice to add a verification:
If `lapiKey|lapiKeyFile` or `appsecKey|appsecKeyFile` is defined, so we avoid cases where none of it was declared directly at startup.  

What do you think ?

pkg/configuration/configuration.go
```diff
@@ -285,8 +289,14 @@ func ValidateParams(config *Config) error {
 	if err != nil {
 		return err
 	}
+
+	// Ensure that a Crowdsec key is provided
+	if appsecKey == "" && lapiKey == "" {
+		return errors.New("CrowdsecLapiKey || CrowdsecAppsecKey: cannot be empty")
```
maxlerebourg: The two key can be empty for mTLS based auth. 

One key must be specified for AppsecMode only, because currently appsec doesn't support mTLS, but in the future, it will be, please remove this test.

PR #313, opened 2026-02-12
Update README.md to show simplified dynamic configuration

cliffjao:
The full dynamic configuration can be pretty intimidating as a starting point. Provide a simplified dynamic configuration to get up & running quickly.


cliffjao (commented):
Updated enabled to true

mathieuHa (approved):

PR #321, opened 2026-04-25
:bug: fix appsec bypass with invalid content-length

maxlerebourg:


mathieuHa (approved):
This looks good to me

HTTP/1.1 allows chunked transfer encoding, where Content-Length is absent or misleading. An attacker can craft a request with a body but no valid Content-Length, and the code before would skip AppSec analysis completely on that body. This could be an AppSec bypass vector.

A bit of context for users:

> An attacker could bypass AppSec inspection entirely by sending a request with a body but setting Content-Length: 0 (or omitting the header entirely, which defaults to -1 in Go's http.Request). Since ContentLength <= 0 was falsy in the condition, the body would never be forwarded to AppSec for analysis, even though it contained potentially malicious payload.

PR #324, opened 2026-04-30
🍱 Header.Add -> Set, log Info -> Debug, version bump

maxlerebourg:
crowdsec doc for header User-Agent for appsec: https://docs.crowdsec.net/docs/next/appsec/protocol


mathieuHa (approved):
Looks good to me

PR #325, opened 2026-05-18
Add parameter to configure the ban page Content-Type response header

highpingblorg:
Defaults to `text/html; charset=utf-8`

Allows users to set their own custom Content-Type response header. This is very useful for Fronted+Backend applications, it allows the backend to serve, for example, a JSON formatted ban page that the Frontend can parse, format and stylize.


mathieuHa (commented):
Two small follow-ups on top of the existing change. Both target edge cases that the YAML-loaded config can produce.

---

**1. Fall back to default when `BanResponseContentType` is empty**

Traefik's plugin loader writes user-supplied YAML *into* the struct *after* `configuration.New()` runs. If an operator explicitly sets `banResponseContentType: ""`, the `New()` default gets overwritten with `""` and `handleBanServeHTTP` ends up calling `rw.Header().Set("Content-Type", "")`, which emits an empty `Content-Type` header.

Cheapest fix: normalize in `ValidateParams` so the rest of the code can trust the value.

In `pkg/configuration/configuration.go`, right after the `BanHTMLFilePath` validation block:

```go
	if config.BanHTMLFilePath != "" {
		if _, err := GetHTMLTemplate(config.BanHTMLFilePath); err != nil {
			return err
		}
	}
	if config.BanResponseContentType == "" {
		config.BanResponseContentType = "text/html; charset=utf-8"
	}
```

---

**2. Set `Content-Type` even when `banTemplate == nil`**

`handleBanServeHTTP` currently skips the header entirely when no template is configured — the early return at `if bouncer.banTemplate == nil` happens before the `Set("Content-Type", ...)`. Moving the header set above the early return applies the configured value uniformly, including for empty-body ban responses (which is useful when the operator wants e.g. `application/json` machine-readable responses without a template).

In `bouncer.go`, replace the relevant block with:

```go
	if bouncer.remediationCustomHeader != "" {
		rw.Header().Set(bouncer.remediationCustomHeader, "ban")
	}
	rw.Header().Set("Content-Type", bouncer.banResponseContentType)
	if bouncer.banTemplate == nil {
		rw.WriteHeader(bouncer.remediationStatusCode)
		return
	}
	rw.WriteHeader(bouncer.remediationStatusCode)
```

---

Other smaller things worth considering (not blockers): validating the value in `ValidateParams` against CR/LF (defense-in-depth, the config is operator-trusted but cheap), mirroring this knob for the captcha/remediation paths, and a README note clarifying that the user also needs a body template matching the configured Content-Type via `BanHTMLFilePath`.

maxlerebourg:
Can we infer the response type from the file the user provide ? A little mapping table of the file extension could avoid 2 new config variables ?

highpingblorg:
> Can we infer the response type from the file the user provide ? A little mapping table of the file extension could avoid 2 new config variables ?

Probably a much cleaner idea. We would then have to manually add support for each unique Content-Type (if someone would need some edge case type).

pkg/configuration/configuration.go
```diff
@@ -396,6 +403,22 @@ func validateCaptcha(config *Config) error {
...
 }
 
+func validateResponseContentTypes(config *Config) error {
+	if config.BanResponseContentType == "" {
+		config.BanResponseContentType = "text/html; charset=utf-8"
+	}
+	if strings.ContainsAny(config.BanResponseContentType, "\r\n") {
+		return errors.New("BanResponseContentType: must not contain CR or LF characters")
+	}
+	if config.CaptchaResponseContentType == "" {
```
mathieuHa: Looks good to me but here do we really need to change the default value in a validation function ?

mathieuHa (changes requested):

pkg/configuration/configuration.go
mathieuHa: @highpingblorg can you look at this ? otherwise everything looks good to me. After we close this comment, i'll ask for final @maxlerebourg approval

mathieuHa: Looks good for this

mathieuHa:
Hi, @highpingblorg, final tests before we merge this!!

maxlerebourg:
Can you merge the main branch.
I really prefer to infer it from BanHTMLFilePath/CaptchaHTMLFilePath but it need a few changes first.
Please add the option to edit your branch.

highpingblorg:
> Can you merge the main branch. I really prefer to infer it from BanHTMLFilePath/CaptchaHTMLFilePath but it need a few changes first. Please add the option to edit your branch.

Done!

maxlerebourg (approved):

bouncer.go
```diff
@@ -184,9 +192,10 @@ func New(_ context.Context, next http.Handler, config *configuration.Config, nam
 		}
 	}
 
-	var banTemplate *htmltemplate.Template
-	if config.BanHTMLFilePath != "" {
-		banTemplate, _ = configuration.GetHTMLTemplate(config.BanHTMLFilePath)
+	var banTemplate *template.Template
+	var banTemplateContentType string
+	if config.BanFilePath != "" {
+		banTemplate, banTemplateContentType, _ = configuration.GetTemplate(config.BanFilePath)
```
mathieuHa, line 198: Here, last argument should be an Err type rigth ?

Could we check return ?

README.md
```diff
@@ -513,14 +513,14 @@ make run
   - int64
   - default: 1800 (= 30 minutes)
   - Period after validation of a captcha before a new validation is required if Crowdsec decision is still valid
-- CaptchaHTMLFilePath
```
mathieuHa, line 516: We could keep the value and add depreciated, or do this

mathieuHa (changes requested):
Look at error return at template build

bouncer.go
mathieuHa: Done in validate

README.md
mathieuHa: Was just a comment

mathieuHa (approved):
OK

PR #329, opened 2026-05-30
✨ tests: CI end-to-end suite — Traefik binary + mock LAPI

### LINKED ISSUE #272: [FEATURE] Improve test coverage

Add e2e test coverage to the project. Unit tests in GO are OK but nothing more robust than full e2e test coverage to ensure that not only the middleware works, but that it does so when consumed from within Traefik itself (plus there are things that you can only test when integrated with Traefik itself such as log output and the interaction with Traefik or Crowdsec).

This also allows us to test Traefik/Crowdsec version upgrades to detect breaking changes in components we depend on.

Something that is cross platform and easy to run locally that spins up a fully working traefik+crowdsec environment to test things like:

* Access logs and headers
* Adding/Removing decisions to crowdsec and these becoming effectively banned by the middleware

**Additional context**

This is in preparation for some upcoming PR's to ensure nothing breaks (plus some of them rely on this kind of test coverage).

mathieuHa:
End-to-end suite that **CI** runs: Traefik as a downloaded **binary** with the plugin loaded from source, and Crowdsec replaced by a small **stdlib-only Go mock LAPI**. No Docker, no real Crowdsec.

> This is the **CI** half of the e2e split suggested in review. The **local** Docker suite (real Traefik + Crowdsec, incl. appsec) lives in #333.

## What runs
- `tests/e2e/mock/mocklapi/` — mock LAPI + stub upstream, in its own nested Go module (excluded from the plugin build/lint/test/vendor).
- `tests/e2e/mock/scenarios/<name>/` — `dynamic.yml` (Traefik file provider) + `run.sh` assertions. Scenarios: `stream-mode`, `live-mode`, `none-mode`, `trusted-ips`, `custom-ban-page`, `captcha`.
- `.github/workflows/e2e.yml` — runs `make -k e2e_mock` on `ubuntu-latest` (sets up Go, downloads the Traefik binary, compiles the mock).

## Scope
Validates the **plugin's own behaviour** (modes, caching, trusted-IP bypass, ban/captcha rendering). Crowdsec / AppSec correctness is **out of scope on purpose** — that is the upstream maintainer's responsibility — so there is no appsec scenario here and the README says so.

## Run locally

```bash
make e2e_mock                 # all scenarios
make e2e_mock_stream-mode     # one scenario
```

Prereqs: `bash`, `curl`, `go`, `tar` (Traefik binary + mock are built/cached on first use).


tests/e2e/mock/mocklapi/main.go
```diff
@@ -0,0 +1,232 @@
...
+		w.WriteHeader(http.StatusOK)
+		if r.Method != http.MethodHead {
+			_, _ = w.Write([]byte("E2E_BACKEND_OK\n"))
+		}
+	})
+}
+
+func main() {
+	lapiAddr := flag.String("lapi-addr", "127.0.0.1:8090", "address for the LAPI mock")
+	backendAddr := flag.String("backend-addr", "127.0.0.1:8091", "address for the backend responder")
```
maxlerebourg: appsecAddr ? or it's an hallucination ?

tests/e2e/mock/.gitignore
```diff
@@ -0,0 +1,2 @@
+# Cached Traefik binary and compiled mock, produced on first run.
+.cache/
```
maxlerebourg: move it in global .gitignore ?

maxlerebourg: We need to download it in each CI no ? The comment above is wrong ?

tests/e2e/mock/mocklapi/main.go
mathieuHa: Not AppSec, and not a hallucination 🙂 — `--backend-addr` is the stub upstream service Traefik proxies *allowed* requests to, i.e. the binary-suite equivalent of the `traefik/whoami` container in the Docker suite. Clarified the flag help text + added a comment in abdb997.

tests/e2e/mock/.gitignore
mathieuHa: Done — moved the `.cache/` rule to the repo root `.gitignore` and removed the per-suite one (abdb997).

mathieuHa: Good catch, the wording was ambiguous. The cache persists across **local** runs, but CI runs on fresh runners so the Traefik binary + mock are indeed recreated on every CI run. Reworded the comment to say exactly that (abdb997).

mathieuHa:
Thanks for the review — three things addressed:

**1. Simplified the mock** (`mocklapi/main.go`: 234 → 118 lines, commit fadd3c3)
- Dropped the stream *delta* bookkeeping (the `startup` flag + "already streamed" set). The plugin re-Sets/Deletes its cache on every poll, so reporting the whole active set as `new` and removed ones as `deleted` is enough.
- Shrank `Decision` to the 3 fields the plugin actually reads (`value`/`type`/`duration`); dropped id/origin/scope/scenario + the id counter.
- Dropped the `X-Api-Key` auth and the `/admin/reset` endpoint — no scenario exercises either.
- Replaced the `store` struct + methods with two package-level maps + a mutex.

**2. On doing the mock with `netcat`/a basic tool** — I looked into it and Id argue against it. The mock isn't serving canned bytes: it needs *mutable state changed mid-test* (`/admin` add/delete drives what `/v1/decisions` returns) and *concurrent* requests (the plugin's stream ticker polls in parallel with live queries + keep-alive). `nc -l` serves one connection at a time, serially, with no routing/parsing/state — so we'd end up hand-rolling an HTTP server in a bash+nc loop (request parsing, Content-Length, a state file with locking, races), which is the classic fragile pattern and *more* complex than what we have. The Go version is already the minimal form of "tiny stateful concurrent HTTP server": stdlib-only, zero deps, same language as the plugin, one binary. So I kept Go but trimmed it hard (point 1). Happy to revisit if you had a specific lighter tool in mind.

**3. The `sleep 4`s** — good catch, they were magic numbers (= `updateIntervalSeconds` + buffer) and not really necessary. Replaced every post-decision `sleep` with a poll on the actual condition (`wait_for_status`, or `wait_for_body_contains` for captcha where the status stays 200 and only the body changes), commit c3b4c3a. This is self-documenting, returns on the first poll that sees the change (faster), and is more robust under slow CI. The full suite now runs in ~22s locally with zero fixed sleeps; control assertions that must *not* change stay immediate.

mathieuHa:
On the Go version divergence: aligned everything on **1.23** — bumped the mock module here, and the plugin `go.mod` in #334 (build/vet/test green on 1.23, one-line diff).

All three PRs (#329, #333, #334) are green and focused now — ready to merge whenever you are, the sooner the better so we stop juggling rebases 😄

tests/e2e/mock/README.md
```diff
@@ -0,0 +1,84 @@
...
+## Scope — what this suite does and does NOT test
+
+These tests validate the **plugin's own behaviour**: the request flow through
+the Traefik middleware, the live / none / stream modes, caching, trusted-IP
+bypass, and ban / captcha page rendering.
+
+They deliberately **do not** test that Crowdsec or its AppSec engine work
+correctly — that is validated and owned by the upstream maintainer
+([@maxlerebourg](https://github.com/maxlerebourg)), not by this plugin. The
+mock only emulates the slice of the LAPI HTTP contract the plugin consumes.
```
maxlerebourg: WTF ?

tests/e2e/mock/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+So please **don't open issues here about Crowdsec/AppSec detection accuracy**
+based on this suite: the AppSec scenario is intentionally absent, and the mock
+returns whatever decisions the test tells it to.
```
maxlerebourg: ??

tests/e2e/mock/scenarios/custom-ban-page/run.sh
```diff
@@ -0,0 +1,24 @@
...
+
+body() {
+  echo "[$SCENARIO] adding ban decision"
+  lapi_add_decision 1.2.3.4 ban 5m
+
+  echo "[$SCENARIO] banned response becomes 403 once the next stream poll lands"
+  wait_for_status "http://127.0.0.1:${WEB_PORT}/foo" 403 15 -H "X-Forwarded-For: 1.2.3.4"
+
+  echo "[$SCENARIO] banned response Content-Type is HTML"
+  assert_header "http://127.0.0.1:${WEB_PORT}/foo" Content-Type "text/html; charset=utf-8" -H "X-Forwarded-For: 1.2.3.4"
```
maxlerebourg, line 18: There is another header to test, via `RemediationHeadersCustomName`

tests/e2e/mock/README.md
mathieuHa: Fixed

tests/e2e/mock/scenarios/custom-ban-page/run.sh
mathieuHa: Fixed

mathieuHa:
I added a new test for appsec with an easy mock with the `rpc2` endpoint from documentation

maxlerebourg (approved):

PR #330, opened 2026-05-30
🤖 chore: adopt Renovate for automated dependency updates

mathieuHa:
This PR now does two things: a one-time version catch-up **and** adopts **Renovate** so these bumps stop being manual.

## 1. Adopt Renovate (replaces Dependabot)

`renovate.json` + a self-hosted weekly workflow (`.github/workflows/renovate.yml`, Mondays 04:00 UTC + manual dispatch). Dependabot is removed (running both would open duplicate PRs).

Renovate covers what Dependabot could not reach here:
- example `docker-compose` image tags (traefik / crowdsec) — native manager;
- the **plugin self-pin** (`experimental.plugins.bouncer.version=` in compose args, `version:` in the Traefik Helm values) — customManager via `github-tags`;
- the Crowdsec/Traefik image tags in the Helm `values.yml` (no `repository` key → matched by file) — customManager via `docker`;
- the pinned Traefik binary in the e2e mock suite (`TRAEFIK_VERSION`) — customManager via `github-releases`;
- `go.mod` + GitHub Actions — native managers (what Dependabot did).

Renovate PRs trigger the existing e2e CI (#329), so a Traefik/Crowdsec bump is validated to actually boot & bounce before merge.

> ⚠️ Requires a `RENOVATE_TOKEN` repo/org secret (PAT) for the self-hosted run — documented in the workflow header. The config can be dry-run via the "Run workflow" button once the secret is set.

## 2. One-time version catch-up

| Component | Before | After |
| --- | --- | --- |
| Traefik image | `v3.0.0` / `v3.5.0` | `v3.7.1` |
| Crowdsec image | `v1.6.1-2` / `v1.6.8` / `latest` | `v1.7.8` |
| Plugin pin | `v1.3.0` / `v1.4.5` / `v1.5.0` | `v1.6.0` |

Covers all `docker-compose.yml` examples + the Helm `values.yml`. From here on, Renovate keeps them current.

## Notes
- `examples/binary-vm/` — no image pin (Vagrant + apt).
- `examples/appsec-enabled/` collections kept as `appsec-virtual-patching` (matches the `/foo/rpc2` CVE demo).
- Same Traefik + Crowdsec versions are exercised green by the e2e scenarios (#329).


maxlerebourg:
set up renovate in scheduled CI for do it automatically

mathieuHa:
@maxlerebourg this PR now also adopts **Renovate** (self-hosted, weekly) to replace the manual version bumps — `renovate.json` + `.github/workflows/renovate.yml`, and Dependabot is removed (running both = duplicate PRs).

It covers what Dependabot could not reach: the example compose image tags, the **plugin self-pin** (compose CLI arg + Helm `version:`), the Helm `values.yml` image tags, the e2e Traefik binary pin, plus go.mod + Actions. Renovate PRs trigger the e2e CI (#329), so a Traefik/Crowdsec bump is validated to boot & bounce before merge.

⚠️ One thing on your side: it needs a **`RENOVATE_TOKEN`** repo/org secret (a PAT with `repo` scope, or fine-grained contents:write + pull-requests:write) for the self-hosted run. Once set, you can dry-run it from the Actions tab → "Renovate" → "Run workflow". Config validation is best confirmed on that first run.

mathieuHa:
### Why the Go floor is pinned to 1.22 (context for the Renovate cap + yaegi pin)

The plugin isn't compiled — Traefik interprets it with **yaegi**, whose bundled stdlib is generated for a specific Go version. That version is the real ceiling for any Go syntax/stdlib the plugin may use.

State of play (checked today):
- Latest **yaegi release is v0.16.1 (Apr 2024)** — its stdlib targets **Go 1.22**; no release since, and `master` is still on the go1.22 stdlib.
- **Every Traefik bundles `yaegi v0.16.1` — including v3.7.1 and `master`** (Traefik's go.mod requires it). Traefik being *built* with Go 1.25 is irrelevant; the embedded interpreter is what runs plugins.
- No bump is in flight: yaegi has an **open** request for Go 1.24 ([traefik/yaegi#1703](https://github.com/traefik/yaegi/issues/1703)) but it is unaddressed, and Traefik `master` has not moved off v0.16.1.

So **Go 1.22 is the plugin's ceiling for the foreseeable future** — declaring/using 1.23 would make the plugin fail to load on ~100% of current Traefik installs.

What we do about it here:
- `renovate.json` caps the go.mod `go`/`toolchain` directive at `<1.23` (Renovate keeps deps/Actions/toolchain current but can't push the plugin past yaegi);
- CI builds/tests on Go 1.22 and keeps `make yaegi_test` @ v0.16.1 as the required guard;
- the e2e PR (#329) follows the same floor, and the standalone 1.23 bump (#334) was dropped.

We raise the cap the day a yaegi release adds newer-Go stdlib **and** Traefik adopts it — worth watching traefik/yaegi#1703 and Traefik's go.mod.

_Sources: [traefik go.mod @ v3.7.1](https://github.com/traefik/traefik/blob/v3.7.1/go.mod) & [@ master](https://github.com/traefik/traefik/blob/master/go.mod) (both require yaegi v0.16.1); [yaegi releases](https://github.com/traefik/yaegi/releases) (latest v0.16.1, "support go1.22"); [yaegi#1703](https://github.com/traefik/yaegi/issues/1703)._

mathieuHa:
### For the record: are Traefik Wasm plugins a way around the yaegi/Go-version ceiling?

**Technically yes, but not a practical one for this plugin — and not worth it just for the Go version.**

**Why Wasm escapes yaegi:** a Wasm plugin is **not interpreted by yaegi**; it is precompiled to a `.wasm` module (TinyGo or Go `wasip1`) and loaded via the **http-wasm** ABI (Traefik = host, plugin = client). So it is not bound to yaegi v0.16.1s Go 1.22 stdlib — you could use newer Go syntax/stdlib (within TinyGo/wasip1 limits).

**Why it is not a realistic workaround here:** it is a different ABI → a **full rewrite**, against a **sandboxed** host model that collides with this plugins core:
- **Redis cache (TCP):** http-wasm exposes no host function for arbitrary TCP → the Redis backend becomes effectively impossible.
- **mTLS to the LAPI:** outbound HTTP from Wasm exists (see `traefik/plugindemowasm-http-call`), but the **host** owns the HTTP client, so fine-grained client-cert / custom-CA control is not guaranteed.
- **Stream mode + shared cache:** the plugin relies on a **background ticker** (periodic decision refresh) and **shared global state** across requests; http-wasm is **per-request** with no long-lived guest goroutines/state → architectural mismatch.
- **TinyGo** has stdlib gaps on `crypto/tls`, `net/http` client, `text/template` (ban/captcha pages), reflection — exactly what the plugin uses. Native `wasip1` is more complete but heavier and still sandboxed.

**Bottom line:** the motivation is weak — the plugin does not need Go 1.23+, so Wasm would trade a non-constraint for a major rewrite with real feature-parity risk (Redis, mTLS, stream). **We stay on yaegi + Go 1.22** (the cap/pin in this PR). Wasm is worth revisiting only as a deliberate re-architecture driven by an independent reason (e.g. Traefik deprecating yaegi, or a real perf win — compiled vs interpreted), not to unlock an unused Go version. Worth keeping on the radar since yaegi has been frozen at v0.16.1 since 2024.

_Sources: [Traefik v3 Wasm deep-dive](https://traefik.io/blog/traefik-3-deep-dive-into-wasm-support-with-coraza-waf-plugin) · [plugindemowasm-http-call](https://github.com/traefik/plugindemowasm-http-call) · [http-wasm http-handler ABI](https://http-wasm.io/http-handler-abi/)._

mathieuHa:
### ✅ Renovate config verified locally — ready to merge

Rebased onto `main`, then ran a full dry-run against the branch:

```bash
LOG_LEVEL=debug GITHUB_COM_TOKEN=… \
  npx renovate --platform=local --dry-run=full
```

**Finding (now fixed in e8ab170):** `config:recommended` ships a default `ignorePaths` that excludes `**/examples/**` and `**/tests/**`. That silently disabled 4 of the 5 custom managers and all the example/helm coverage this PR adds — Renovate only saw the root `docker-compose.yml`, `go.mod` and the workflows (7 package files). Fixed by overriding `ignorePaths` to keep only `vendor`/`node_modules`.

**Dry-run after the fix** — 32 package files detected, every custom manager extracts its dep:

| Custom manager | Matches | Extracted |
| --- | --- | --- |
| Plugin self-pin (compose args) | 11 compose files | `maxlerebourg/crowdsec-bouncer-traefik-plugin` |
| Plugin self-pin (Helm values) | `traefik/values.yml` | `v1.6.0` |
| Traefik image (Helm values) | `traefik/values.yml` | `traefik v3.7.1` (docker) |
| Crowdsec image (Helm values) | `crowdsec/values.yml` | `crowdsecurity/crowdsec v1.7.8` (docker) |
| e2e Traefik binary | `tests/e2e/mock/lib/common.sh` | `traefik/traefik v3.7.1` (github-releases) |

Native managers also verified: `traefik/whoami` correctly disabled, Go held at `1.22` by the `<1.23` yaegi cap, Traefik/Crowdsec grouping applied, GitHub Actions picked up. `renovate-config-validator` passes with 0 errors.

Everything behaves as intended — good to merge. 🚀

mathieuHa:
@maxlerebourg — reworked this into a **clean "adopt Renovate" PR**. 

I dropped the one-time manual version catch-up (reverted all 13 example / root `docker-compose` / Helm `values.yml` files back to `main`), since Renovate now owns those bumps. The diff vs `main` is now Renovate-only:

- `renovate.json` + `.github/workflows/renovate.yml` (self-hosted, weekly)
- Dependabot removed
- `main.yml` Go pin (yaegi 1.22 ceiling)

Verified with a local dry-run (`renovate --platform=local --dry-run=full`) after the revert — **32 package files detected**, and Renovate would immediately open PRs to bring the examples up to date:

| Dep | From | To |
| --- | --- | --- |
| traefik (compose/helm) | v3.0.0 / v3.5.0 | v3.7.6 |
| crowdsec (compose/helm) | v1.6.1 / v1.6.8 | v1.7.8 |
| plugin self-pin | v1.3.0 / v1.4.5 / v1.5.0 | v1.6.0 |
| e2e Traefik binary (`common.sh`) | v3.7.1 | v3.7.6 |

It even catches `redis` and `ubuntu` tags the manual pass had missed.

⚠️ One thing to decide: `examples/tls-auth/docker-compose.yml` is back on `crowdsecurity/crowdsec:latest`. Renovate does **not** pin rolling tags, so that one stays floating. If you'd rather have it pinned+managed, say the word and I'll pin it in this PR.

Also note the workflow still needs the `RENOVATE_TOKEN` secret before the scheduled/dispatch run can push PRs.

mathieuHa:
### Why a few crowdsec image tags are being changed

While verifying that Renovate would actually keep the examples current, I hit two tags it **cannot** manage — so I normalized them (this is the only content change left in an otherwise config-only PR):

1. **`examples/tls-auth` was on `crowdsecurity/crowdsec:latest`** → pinned to `v1.6.8`.
   Rolling `latest` isn't reproducible and Renovate never proposes updates for it (there's no version to compare).

2. **8 examples were on `crowdsecurity/crowdsec:v1.6.1-2`** → normalized to `v1.6.8` (+ the captcha README snippet, to match).
   Renovate's docker versioning won't advance a **build-suffixed** tag (`…-2`) to a clean release, so those files would have silently stayed frozen on 1.6.1 forever.

`v1.6.8` is not a new version — it's the tag already used in the root `docker-compose.yml` / `.local.yml`, so this just makes all examples consistent on a tag Renovate understands.

**Verified** (`renovate --platform=local --dry-run=full`): all **11** crowdsec image references are now Renovate-managed and would bump `v1.6.8 → v1.7.8` (0 stuck). Everything else is unchanged — no manual version bumps, Renovate owns them going forward.

renovate.json
```diff
@@ -0,0 +1,72 @@
...
+    {
+      "description": "Group everything Traefik (image, helm tag, e2e binary) into one PR",
+      "matchPackageNames": ["traefik", "traefik/traefik"],
+      "groupName": "Traefik"
+    },
+    {
+      "description": "Group Crowdsec (image + helm tag) into one PR",
+      "matchPackageNames": ["crowdsecurity/crowdsec"],
+      "groupName": "Crowdsec"
+    },
```
maxlerebourg: groupName: all at the root

renovate.json
```diff
@@ -0,0 +1,72 @@
...
+      "datasourceTemplate": "docker"
+    },
+    {
+      "description": "Crowdsec image tag in the Crowdsec Helm values (no repository key, so match by file)",
+      "customType": "regex",
+      "managerFilePatterns": ["/^examples/kubernetes/crowdsec/values\\.ya?ml$/"],
+      "matchStrings": ["tag:\\s*(?<currentValue>v[0-9]+\\.[0-9]+\\.[0-9]+)"],
+      "depNameTemplate": "crowdsecurity/crowdsec",
+      "datasourceTemplate": "docker"
+    },
```
maxlerebourg, line 63: Renovate don't do it via an extends like ":kubernetes" or other ? I don't think a custom manager is needed here

examples/appsec-enabled/docker-compose.yml
```diff
@@ -47,7 +47,7 @@ services:
       - "traefik.http.middlewares.crowdsec.plugin.bouncer.crowdsecappsechost=crowdsec:7422"
 
   crowdsec:
-    image: crowdsecurity/crowdsec:v1.6.1-2
+    image: crowdsecurity/crowdsec:v1.6.8
```
maxlerebourg, line 50: can we revert and let renovate open the next PR with these automatic change

examples/tls-auth/docker-compose.yml
```diff
@@ -71,7 +71,7 @@ services:
       # Define AppSec host and port informations
       - "traefik.http.middlewares.crowdsec.plugin.bouncer.crowdsecappsechost=crowdsec:7422"
   crowdsec:
-    image: crowdsecurity/crowdsec:latest
+    image: crowdsecurity/crowdsec:v1.6.8
```
maxlerebourg, line 74: ok for this one

renovate.json
mathieuHa: what do you mean?

renovate.json
mathieuHa: I think it is. Values is yaml and can contain anything in any format. So when updating docker image, you need to say where it is and even which image to renovate

maxlerebourg (approved):

renovate.json
```diff
@@ -2,25 +2,20 @@
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": ["config:recommended"],
   "labels": ["dependencies"],
-  "prHourlyLimit": 0,
   "ignorePaths": ["**/vendor/**", "**/node_modules/**"],
+  "rangeStrategy": "bump",
+  "prConcurrentLimit": 1,
+  "branchPrefix": "renovate/",
+  "commitMessagePrefix": "⬆️ renovate: ",
+  "groupName": "all",
```
mathieuHa, line 10: If that does what I think we can't bump traefik apart from crowdsec with that setting

PR #331, opened 2026-05-30
:bug: fall back to system trust store when no custom TLS CA is set

### LINKED ISSUE #327: [FEATURE] Support CrowdsecLapiHost with a valid certificate
This is a feature request.
I'm running the CrowdSec LAPI on a host behind a reverse proxy under `crowdsec.example.com`, which means the LAPI API has a valid Let's Encrypt certificate.

The Traefik Bouncer currently connects to the LAPI using this configuration, which works fine:
```yaml
# traefik.yml
[...]
http:
  middlewares:
    crowdsec-bouncer:
      plugin:
        bouncer:
          Enabled: true
          CrowdsecMode: stream
          CrowdsecAppsecEnabled: true
          CrowdsecAppsecScheme: http
          CrowdsecLapiScheme: https
          CrowdsecLapiHost: crowdsec.example.com
          CrowdsecLapiTLSInsecureVerify: true
          CrowdsecLapiKey: <LAPI-API-Key>
[...]
```

The problem is that I always have to specify either `CrowdsecLapiTlsInsecureVerify` or `CrowdsecLapiTlsCertificateAuthority` so that the plugin doesn’t throw an error. This means that in this case, I currently have no choice but to trust insecure certificates, since without both options, an error is thrown. 

It should be possible to specify neither `CrowdsecLapiTlsInsecureVerify` nor `CrowdsecLapiTlsCertificateAuthority`.

mathieuHa:
Closes #327.

## Problem

Setting `crowdsecLapiScheme=https` against a LAPI exposed behind a reverse proxy with a publicly trusted certificate (e.g. Let's Encrypt) currently forces the operator to pick one of two suboptimal options:

- set `crowdsecLapiTLSInsecureVerify=true` (skip verification — defeats the point), or
- pin `crowdsecLapiTLSCertificateAuthority` to a CA bundle (extra config nobody should need for a public CA).

The expected behavior — "use the system trust store" — wasn't reachable.

## Root cause

Two contributing bugs in `pkg/configuration/configuration.go`:

1. `validateParamsTLS` returned an error when `CrowdsecLapiTLSCertificateAuthority` was empty.
2. `getTLSConfig` always initialized `tls.Config.RootCAs = x509.NewCertPool()` (an empty pool), which silently *disables* the stdlib's fall-back to `x509.SystemCertPool()`. So even if validation didn't fire, verification of a Let's Encrypt cert would still fail.

The Go idiom for "use the system trust store" is to leave `tls.Config.RootCAs` as `nil`. The `traefik` image (alpine + `ca-certificates`) ships a CA bundle that already trusts the public roots.

## Fix

- `validateParamsTLS`: drop the empty-CA error; still rejects garbage PEM.
- `getTLSConfig`: only allocate `RootCAs` when a custom CA is provided; otherwise leave it `nil`.
- Same change applies symmetrically to the AppSec path (helper is shared).

## Tests

`Test_validateParamsTLS` and `Test_GetTLSConfigCrowdsec` were scaffolds (`// TODO: Add test cases.`) — now filled in.

`Test_validateParamsTLS` covers:
- Empty CA → accepted
- Valid PEM → accepted
- Garbage CA → rejected

`Test_GetTLSConfigCrowdsec` covers:
- HTTP scheme → empty `tls.Config`
- HTTPS + no CA → `RootCAs == nil` (system trust store)
- HTTPS + custom CA → `RootCAs` populated
- HTTPS + insecure verify → `InsecureSkipVerify == true`
- HTTPS + garbage CA → error

## Docs

`README.md` — replaced the "you need to either accept any certificates or add the CA" wording with three explicit options (system trust store / custom CA / skip verification), for both the LAPI and the AppSec sections. Per-field descriptions of `Crowdsec{Lapi,Appsec}TlsCertificateAuthority` now mention the system trust store as the default fallback.

## Compatibility

No behavior change for existing setups: configs that explicitly set a custom CA or insecure verify continue to work identically. The new behavior is only reachable by *omitting* both — which was rejected before.

### DIFF
--- a/pkg/configuration/configuration_test.go
+++ b/pkg/configuration/configuration_test.go
@@ -6,6 +6,20 @@ import (
 	logger "github.com/maxlerebourg/crowdsec-bouncer-traefik-plugin/pkg/logger"
 )
 
+// validPEM is a minimal self-signed certificate accepted by AppendCertsFromPEM,
+// shared by the TLS tests below.
+const validPEM = `-----BEGIN CERTIFICATE-----
+MIIBhTCCASugAwIBAgIQIRi6zePL6mKjOipn+dNuaTAKBggqhkjOPQQDAjASMRAw
+DgYDVQQKEwdBY21lIENvMB4XDTE3MTAyMDE5NDMwNloXDTE4MTAyMDE5NDMwNlow
+EjEQMA4GA1UEChMHQWNtZSBDbzBZMBMGByqGSM49AgEGCCqGSM49AwEHA0IABD0d
+7VNhbWvZLWPuj/RtHFjvtJBEwOkhbN/BnnE8rnZR8+sbwnc/KhCk3FhnpHZnQz7B
+5aETbbIgmuvewdjvSBSjYzBhMA4GA1UdDwEB/wQEAwICpDATBgNVHSUEDDAKBggr
+BgEFBQcDATAPBgNVHRMBAf8EBTADAQH/MCkGA1UdEQQiMCCCDmxvY2FsaG9zdDo1
+NDUzgg4xMjcuMC4wLjE6NTQ1MzAKBggqhkjOPQQDAgNIADBFAiEA2zpJEPQyz6/l
+Wf86aX6PepsntZv2GYlA5UpabfT2EZICICpJ5h/iI+i341gBmLiAFQOyTDT+/wQc
+6MF9+Yw1Yy0t
+-----END CERTIFICATE-----`
+
 func getMinimalConfig() *Config {
 	cfg := New()
 	cfg.CrowdsecLapiKey = "test"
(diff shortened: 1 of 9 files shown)


pkg/configuration/configuration_test.go
```diff
@@ -126,19 +124,37 @@ func Test_ValidateParams(t *testing.T) {
...
+DgYDVQQKEwdBY21lIENvMB4XDTE3MTAyMDE5NDMwNloXDTE4MTAyMDE5NDMwNlow
+EjEQMA4GA1UEChMHQWNtZSBDbzBZMBMGByqGSM49AgEGCCqGSM49AwEHA0IABD0d
+7VNhbWvZLWPuj/RtHFjvtJBEwOkhbN/BnnE8rnZR8+sbwnc/KhCk3FhnpHZnQz7B
+5aETbbIgmuvewdjvSBSjYzBhMA4GA1UdDwEB/wQEAwICpDATBgNVHSUEDDAKBggr
+BgEFBQcDATAPBgNVHRMBAf8EBTADAQH/MCkGA1UdEQQiMCCCDmxvY2FsaG9zdDo1
+NDUzgg4xMjcuMC4wLjE6NTQ1MzAKBggqhkjOPQQDAgNIADBFAiEA2zpJEPQyz6/l
+Wf86aX6PepsntZv2GYlA5UpabfT2EZICICpJ5h/iI+i341gBmLiAFQOyTDT+/wQc
+6MF9+Yw1Yy0t
+-----END CERTIFICATE-----`
+
```
maxlerebourg: you use twice the same var, declare it outside, or in a func getPEM()

pkg/configuration/configuration_test.go
```diff
@@ -126,19 +124,37 @@ func Test_ValidateParams(t *testing.T) {
...
+	cfgEmpty := getMinimalConfig()
+	cfgValid := getMinimalConfig()
+	cfgValid.CrowdsecLapiTLSCertificateAuthority = validPEM
+	cfgGarbage := getMinimalConfig()
```
maxlerebourg: cfgGarbage, name it seriously please 🥇

pkg/configuration/configuration_test.go
mathieuHa: Fixed in b8fefb3 — the duplicated `validPEM` block is now a single package-level const, both local copies removed.

pkg/configuration/configuration_test.go
mathieuHa: Fixed in b8fefb3 — renamed `cfgGarbage` → `cfgInvalidCA` (and the matching test case).

maxlerebourg (dismissed):

maxlerebourg (approved):

PR #333, opened 2026-06-05
✨ tests: local Docker end-to-end suite (Traefik + Crowdsec)

mathieuHa:
Local-only end-to-end suite that boots **real Traefik + Crowdsec** containers and exercises the plugin loaded from a local path.

This is the **local** half of the e2e split suggested in review — the **CI** half (Traefik binary + mock LAPI) lives in #329.

## Scope
- `tests/e2e/scenarios/<name>/` — one isolated Docker Compose stack per scenario.
- Scenarios: `stream-mode`, `live-mode`, `none-mode`, `trusted-ips`, `custom-ban-page`, `captcha`, `appsec`.
- `tests/e2e/lib/common.sh` — shared curl/status/header helpers + diagnostics dump.

## Why local-only (not in CI)
It boots a real Crowdsec and, for `appsec`, downloads the OWASP CRS collections — heavier and less deterministic than CI needs. CI runs the lighter binary + mock-LAPI suite (#329) instead. This suite stays available for high-fidelity local debugging against a real Crowdsec.

## Run

```bash
make e2e            # all scenarios
make e2e_stream-mode
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,8 @@
-.PHONY: lint test vendor clean
+.PHONY: lint test vendor clean e2e
 
 export GO111MODULE=on
+
+E2E_SCENARIOS := stream-mode live-mode none-mode trusted-ips custom-ban-page captcha appsec
 
 default: lint test
 
@@ -12,6 +14,11 @@ test:
 
 yaegi_test:
 	yaegi test -v .
+
+e2e: $(addprefix e2e_,$(E2E_SCENARIOS))
+
+e2e_%:
+	./tests/e2e/scenarios/$*/run.sh
 
 vendor:
 	go mod vendor
--- a/tests/e2e/README.md
+++ b/tests/e2e/README.md
@@ -1,0 +1,55 @@
+# End-to-end test suite
+
+These tests spin up real Traefik + Crowdsec containers and exercise the
+plugin in the same conditions Traefik uses in production: loaded from a
+local path, no module download, no mocking.
+
+Each scenario lives in its own directory under `scenarios/` and owns:
+
+- `docker-compose.yml` — the stack to spin up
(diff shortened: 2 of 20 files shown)


david-garcia-garcia:
@maxlerebourg it is going to be difficult to survive the speed of the AI era without truly e2e tests that test 3d party dependencies without mocking them. If someone forks with proper AI testing and guardrails, they will superseed the relibaility and functionality of this plugin in a matter of weeks.

This thing only exists as a Traefik + CrowdSec plugin. If those two move and we only test a mock we wrote ourselves, CI stays green and users are the ones who notice.

I’m not saying drop the mock suite. It’s fast and it’s useful. It just can’t tell you whether real stream, cscli, AppSec, or Traefik’s plugin loader still work. That’s a photo of the API we invented, not the one they ship.

“That’s upstream’s problem” is fair for CRS rules. It’s not fair for how we call them. And a Docker suite that only runs on someone’s laptop might as well not exist. Pins drift, nobody runs it on the PR.

This gets worse with AI, not better. Models treat CI as the spec. If the only required check is our mock, they’ll keep the mock happy and never pull a real image. That’s exactly when a Traefik or CrowdSec bump breaks us.

So: keep the mock, and put a real Traefik + CrowdSec job on the PR. Pinned images, cscli for decisions, one AppSec case against the actual engine. If it’s optional, it won’t run.

 @mathieuHa 
 
 > This suite stays available for high-fidelity local debugging against a real Crowdsec.
 
 If this was up to me, this would run on every CI, even at the expense of having glitches unrelated to the crowdsec bouncer itself.

I already thought about this before AI kicked the hornet’s nest: https://github.com/maxlerebourg/crowdsec-bouncer-traefik-plugin/issues/272

PR #334, opened 2026-06-05
⬆️ chore: bump go.mod to go 1.23

mathieuHa:
Align the module's declared Go version with what the project actually uses.

## Why
- `go.mod` declared `go 1.22`, but CI (`main.yml`) already builds and tests on **1.23**, and the README "go.mod Go version" badge advertises whatever `go.mod` says. So the advertised/declared minimum (1.22) was never the version CI exercised — a silent divergence that keeps drifting.
- This bumps the declared version to **1.23** so CI tests the advertised minimum and everything lines up on one version.

## Verification (Go 1.23, `GOTOOLCHAIN=local`)
- `go build ./...` ✅
- `go vet ./...` ✅
- `go test ./...` ✅
- `go mod verify` → all modules verified ✅
- One-line diff, no `toolchain` line added, no vendor change.

The README badge will reflect 1.23 automatically once merged.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/maxlerebourg/crowdsec-bouncer-traefik-plugin
 
-go 1.22
+go 1.23
 
 require (
 	github.com/leprosus/golang-ttl-map v1.1.7


mathieuHa:
Closing this: after digging into the Traefik/yaegi compatibility, bumping the plugin to Go 1.23 is **not safe**.

The plugin is interpreted by **yaegi**, and even the latest Traefik (v3.7.1) bundles **yaegi v0.16.1, which only supports Go 1.22** (Traefik itself is built with Go 1.25, but that does not matter — the embedded yaegi is what interprets the plugin). Declaring `go 1.23` would advertise a minimum the interpreter cannot honour, and any Go 1.23 stdlib/language feature would break the plugin on **every** current Traefik install.

So we keep the plugin on **Go 1.22** and add a Renovate cap (`<1.23`) tied to yaegi, plus keep `yaegi_test` (v0.16.1) as the required guard. We can raise the floor the day Traefik ships a yaegi with 1.23 support. Details in #330.

Issue #339, opened 2026-06-17
[FEATURE] Add pass through for captcha-pending IPs requesting custom captcha resources


matthesvoss:
**Is your feature request related to a problem? Please describe.** 🐛
When using a custom captcha like wicketkeeper and an IP is captcha-flagged, it is not able by default to request the necessary resources to complete the captcha (/fast.js and /v0/challenge). Instead it is prompted with another captcha challenge html.

**Describe the solution you'd like** ✨
1. A new CaptchaCustomChallengeURL configuration option should be added for /v0/challenge.
2. Captcha-flagged IPs should be passed through to CaptchaCustomJsURL and CaptchaCustomChallengeURL. Banned IPs should still not be passed through.

samtoxie:
this would be a good QOL improvement indeed, faced the same error and just ended up serving wicketkeeper on a seperate entrypoint listening on 9443 without the bouncer. Not the prettiest but it works, but bypass would indeed be better.

mathieuHa:
Hi, we'll go over recently opened issues / MR shortly we just finished a merge windows that landed many fixes/improvements (1.7.0).  Will probably do that in next version!!

PR #340, opened 2026-06-17
Add pass through for captcha-pending IPs requesting custom captcha resources

matthesvoss:
This PR solves issue #339.


mathieuHa:
Just noting some AI review for indication, I need a manual pass on this.

```
Review of PR #340 — "Add pass through for captcha-pending IPs requesting custom captcha resources"

  The PR fixes #339: with a self-hosted (custom) captcha provider behind the bouncer, captcha-pending IPs couldn't fetch the widget's JS/challenge endpoints, so the captcha could never be solved. It adds a
  CaptchaCustomChallengeURL option, parses it plus CaptchaCustomJsURL into match targets, and lets captcha-pending requests to those exact host+path combinations pass through to the next handler. The core
  wiring is correct — the new New() argument order matches at the single call site, pass-through still goes through appsec, banned IPs remain blocked, and the check sits correctly after the grace-period
  shortcut.

  Findings, most severe first:

  1. pkg/captcha/captcha.go:214 — Host matching does no port normalization, silently defeating the feature for common config styles (configure http://example.com:80/fast.js and the browser sends Host:
  example.com → "example.com:80" != "example.com" → no pass-through, no warning, and the #339 captcha loop returns; same failure in reverse when the config omits a port but the deployment runs on a
  non-standard one).
  2. pkg/captcha/captcha.go:42 — Invalid config values are only downgraded to a runtime Warn instead of failing startup validation like every other URL option (a scheme-less typo such as
  captcha.localhost:8000/fast.js parses with Path == "" — Go treats captcha.localhost as the scheme — as does a root-URL value like http://captcha.localhost:8000; the entry is dropped and pass-through is
  silently disabled, while validateCaptcha in pkg/configuration/configuration.go:416 is the established place to reject this at startup).
  3. pkg/captcha/captcha.go:217 — A path-only configured URL matches that path on every host sharing the middleware instance (e.g. captchaCustomJsURL=/fast.js with the same crowdsec@docker middleware
  attached to several routers lets a captcha-pending IP reach /fast.js on an unrelated vhost, uncounted in blockedRequests; documented in the code comment, but worth a README warning to prefer full URLs
  with a host).
  4. examples/custom-captcha/docker-compose.yml:96 — The example never exercises the feature it documents (the router-wicketkeeper router carries no middlewares=crowdsec@docker label, so requests to
  captcha.localhost bypass the bouncer entirely — the example neither reproduces #339 nor validates the new pass-through; attach the middleware to that router to make the example demonstrate the fix).
  5. bouncer.go:475 — HEAD requests are inconsistently excluded from the pass-through (the pre-existing req.Method != http.MethodHead guard routes a captcha-pending IP's HEAD /fast.js to the ban response
  while GET /fast.js passes through; pre-existing behavior, but the new feature makes the inconsistency visible).
  6. pkg/captcha/captcha.go:84 — Design/altitude: two hardcoded resource URLs and a growing positional signature (New() is now 14 positional parameters, mostly adjacent strings the compiler can't
  distinguish, and the next provider asset — CSS, images, audio challenge — forces yet another config field threaded through three files; a single list option like captchaCustomPassThroughURLs plus a config
  struct for New() would absorb future cases without code changes).
  7. pkg/captcha/captcha.go:189 — No unit tests for the new pure functions (parseCaptchaResourceURLs and IsCaptchaResource are ideal table-test targets — host/port variants, path-only URLs, scheme-less
  typos — and sibling packages like pkg/configuration already have test files establishing the convention).

  Overall: the approach is sound and the plumbing is correct; I'd ask for #1 and #2 (both silently re-break the exact scenario the PR fixes, just with different configs) plus the example fix in #4 before
  merging. #3 deserves at least a docs note, and #5–7 are take-or-leave polish.
```

matthesvoss:
I implemented most of the AI's findings and would appreciate a review.

maxlerebourg:
Instead of adding 3 URLs to the whitelist, perhaps we could add a CAPTCHA domain or a list of authorised URLs.
Many local CAPTCHA providers will need to load more than 3 URLs.

This feature can be separated from the CAPTCHA and offer far more possibilities if the crowdsec-bouncer-plugin is deployed at the endpoint layer.

Tell me if you don’t understand my point of view.
Thanks for the interest in our plugin and your work on this PR.

mathieuHa:
We'll work on this more global whitelist feature in a next PR, once we have merged all in line for 1.8 release.

PR #342, opened 2026-06-22
feat: Allow cache reading from replicas

samtoxie:
Hello, I run Redis in clustered mode with 2 read replicas besides a master.

It would be nice to be able to offload reading to these replicas too, so that the master doesnt get all the load or when it gets restarted (we dont run in sentinel mode). As such I implemented this PR. It is implemented in a way that the read host config is optional, so no breaking changes.


samtoxie:
Hi @maxlerebourg or @mathieuHa,

Any chance on getting this small PR checked and merged for the next release? It would significantly improve QOL for the various clusters I have to manage. :smile:

maxlerebourg (approved):

maxlerebourg (dismissed):

mathieuHa:
Some comment issue in the tests fixed in 8eceba35b7e7a3e4c9edf39508f51d8fc14e1920

mathieuHa:
Added a mock of the read replica feature, since it's kinda introduced in this PR

mathieuHa:
Max what do you think. 
When all replicas are down, should it fallback to the primary for reads or should it fail?

My opinion is we can do both but it would be the user responsability to set the primary in the primary and read replica pool or use some autodiscovery engine that register it automatically

In that can maybe we add some lines of docs on it

mathieuHa:
Last thing, during the rewrite of the function that access a key in redis using simpleredis we found a "bug"

```go
func (rc *redisCache) get(key string) (string, error) {
        value, err := rc.nextReader().Get(key)
        if err == nil {
                valueString := string(value)
                if len(valueString) > 0 {
                        return valueString, nil
                }
        }
        switch err.Error() {           // ← unsafe if err == nil
        case simpleredis.RedisMiss:
                return "", errors.New(CacheMiss)
        case simpleredis.RedisUnreachable:
                return "", errors.New(CacheUnreachable)
        }
        return "", err
  }
  ```

  The panic path. Trace what happens when Get returns err == nil and an empty value:

  1. if err == nil is true, so we enter the block.
  2. len(valueString) > 0 is false, so we don't return.
  3. Execution falls out of the if and reaches switch err.Error().
  4. err is nil. error is an interface, and calling a method on a nil interface value dereferences a nil pointer → runtime panic: "invalid memory address or nil pointer dereference."

In practice data given should never be with len(str)=0 since the plugin uses non-empty verdicts ("t"/"f"/"c"/"d").  
However I think it's better to rearrange error checking to work without that non empty guaranty.

last commit fixes that.

mathieuHa:
Max please review above and then we should be ready to merge

maxlerebourg (approved):

maxlerebourg (dismissed):

mathieuHa (dismissed):

maxlerebourg (approved):

PR #332, opened 2026-06-04
:bug: appsec: do not buffer unreadable (gRPC/HTTP2) request bodies

mathieuHa:
## Problem

Fixes #323.

After #321, AppSec silently returns `403` on long-lived gRPC streams (e.g. NetBird `ConnectStream`) when AppSec is enabled.

A bidirectional gRPC stream is an **HTTP/2 request with no `Content-Length` whose body never reaches EOF**. #321 removed the `httpReq.ContentLength > 0` guard in `appsecQuery`, so the AppSec path now buffers every body with `io.ReadAll`. For an open stream that read blocks until the request times out, producing a `403` **before the backend is reached** (`OriginStatus:0`). Short unary gRPC calls succeed; only streams hang.

## Fix

Mirror the reference `lua-cs-bouncer`, which refuses to read the body of an HTTP/2+ request with no `Content-Length` and then either forwards headers-only or drops it (`APPSEC_DROP_UNREADABLE_BODY`).

- `isBodyUnreadable(req)` → `req.Body != nil && req.ProtoMajor >= 2 && req.ContentLength < 0` (the Go equivalent of the lua `http_version() >= 2 and http_content_length == nil` check).
- When the body is unreadable, `appsecQuery` no longer buffers it:
  - default → forward to AppSec **with headers only** (IP / URI / verb / headers still inspected), request proceeds;
  - new `CrowdsecAppsecDropUnreadableBody` option (default `false`, mirrors `APPSEC_DROP_UNREADABLE_BODY`) → block the request outright.
- Readable HTTP/1.1 bodies are still buffered and inspected → the bypass closed by #321 stays closed.

## Tests

- `Test_isBodyUnreadable` — HTTP/2 & HTTP/3 without Content-Length (true), HTTP/2 with Content-Length (false), HTTP/1.1 chunked (false), no body (false).
- `Test_appsecQuery_streamingDoesNotBlock` — regression for #323: a never-EOF gRPC body returns within 2s (hangs ~forever / 2s timeout without the fix).
- `Test_appsecQuery_dropUnreadableBody` — verifies drop mode blocks.

`go test ./...`, `go vet ./...` and `yaegi test -v .` (Traefik plugin runtime) all pass.

## Docs

`CrowdsecAppsecDropUnreadableBody` documented in the variables list and the full dynamic-configuration example.

Issue #323, opened 2026-04-30
[BUG] Silently returns 403 on gRPC streaming connections


bhz0u:
**Describe the bug** 🐛
After upgrading from v1.5.1 to v1.6.0, the plugin breaks gRPC connections that was previously working. First saw this behavior with self hosted instance of NetBird, where all connections to "/signalexchange.SignalExchange/ConnectStream" gets a DownstreamStatus 403. Reverting to v1.5.1 fixes this problem, no other changes were made to NetBird or Traefik. 

**Expected behavior** 👀
gRPC connections should pass through the bouncer without interference when no Crowdsec decision exists for the client IP.

**Context** 🔎
The plugin is applied as an entrypoint-level middleware on the `websecure` entrypoint, so applied to all HTTPs requests. Traefik terminates TLS and forwards gRPC via `h2c` to the backend.
NetBird uses long-lived gRPC requests to the "/signalexchange.SignalExchange/ConnectStream" endpoint. 
Crowdsec logs no block actions with "cscli decisions list", even a manual allow for clients does not work. (via cscli decisions add --type allow)

Other Observations
- `OriginStatus: 0` and `OriginDuration: 0` shows the request never reached the backend. Crowdsec plugin returned the 403 itself.
- The same client IP successfully completes `GetServerKey` and `Login` gRPC calls (both return `Grpc-Status: 0`) moments before the `ConnectStream` call is blocked.
- `cscli decisions list` shows no decisions for the affected IPs. No logs from this plugin than "msg=handleStreamCache:updated" in Traefik container. 
- No ban/block entries appear in plugin logs even at `DEBUG` level.
- WebSocket connections and standard HTTPS work fine through the same entrypoint. 

Anonymized section of access logs:
```json
# Plugin version v1.6.0
[...]

PR #352, opened 2026-07-10
Do not consider body unreadable when it's http.NoBody

### LINKED ISSUE #351: [BUG] crowdsecAppsecUnreadableBodyBlock should default to false
**Describe the bug** 🐛
Just upgraded the bouncer to latest main, with Traefik 3.7.7. Every GET queries to any service protected by the bouncer middleware started to get a 403. In debug mode I could see
```
level=DEBUG msg="handleNextServeHTTP ip:192.168.7.106 isWaf:true appsecQuery:unreadableBody dropped" component=CrowdsecBouncerTraefikPlugin
```
I guess it's a side effect of https://github.com/maxlerebourg/crowdsec-bouncer-traefik-plugin/commit/d32f271195754703f9393b4d8830c7e4b9da6655

I wouldn't expect it to block absolutely all my GET queries (using http3 between my browser and Treafik). And the doc mention crowdsecAppsecUnreadableBodyBlock should default to false. But it started working again only after I explicitely set it to false

**Expected behavior** 👀
The bouncer shouldn't deny any request. And crowdsecAppsecUnreadableBodyBlock should default to false (but as it forbid every requests when set to true, it has no real interest :-) )

**Version (please complete the following information):**
 - OS: Docker/Alpine
 - Traefik version: 3.7.7
 - Plugin version: current main

dani:
GET requests (or any req which has no body) do not have a nil httpReq.Body on the server side. Instead, its value is http.NoBody. Check for Body == http.NoBody in isBodyUnreadable to prevent blocking every GET request

Fix #351

### DIFF
--- a/bouncer.go
+++ b/bouncer.go
@@ -752,7 +752,6 @@ func appsecQuery(bouncer *Bouncer, ip string, httpReq *http.Request) error {
 		Path:   bouncer.appsecPath,
 	}
 	var req *http.Request
-	bouncer.log.Info("coucou %d", httpReq.ProtoMajor)
 	switch {
 	case isBodyUnreadable(httpReq):
 		if bouncer.appsecUnreadableBodyBlock && isMethodWithBody(httpReq.Method) {
--- a/bouncer.go
+++ b/bouncer.go
@@ -735,11 +735,7 @@ func isBodyUnreadable(httpReq *http.Request) bool {
 	return httpReq.Body != nil && httpReq.Body != http.NoBody && httpReq.ProtoMajor >= 2 && httpReq.ContentLength < 0
 }
 
-// isMethodWithBody reports whether the method is expected to carry a request
-// body. Only those methods may be dropped when their body is unreadable,
-// mirroring the reference lua-cs-bouncer METHODS_WITH_BODY list. Other methods
-// (e.g. a browser GET over HTTP/3, which never carries a Content-Length) are
-// forwarded to the Appsec component headers-only instead of dropped (issue #351).
+// isMethodWithBody used only when isBodyUnreadable returns true but the request method can't have body.
 func isMethodWithBody(method string) bool {
 	switch method {
 	case http.MethodPost, http.MethodPut, http.MethodPatch, http.MethodDelete:
@@ -756,10 +752,10 @@ func appsecQuery(bouncer *Bouncer, ip string, httpReq *http.Request) error {
 		Path:   bouncer.appsecPath,
 	}
 	var req *http.Request
+	bouncer.log.Info("coucou %d", httpReq.ProtoMajor)
 	switch {
 	case isBodyUnreadable(httpReq):
 		if bouncer.appsecUnreadableBodyBlock && isMethodWithBody(httpReq.Method) {
-			// The caller (handleNextServeHTTP) logs this returned error with the IP.
 			return errors.New("appsecQuery:unreadableBody dropped")
 		}
 		req, _ = http.NewRequest(http.MethodGet, routeURL.String(), nil)
--- a/bouncer.go
+++ b/bouncer.go
@@ -732,7 +732,7 @@ func crowdsecQuery(bouncer *Bouncer, stringURL string, data []byte) ([]byte, err
 // a 403. This mirrors the reference lua-cs-bouncer behavior, which refuses to
 // read the body of an HTTP/2+ request that has no Content-Length.
 func isBodyUnreadable(httpReq *http.Request) bool {
-	return httpReq.Body != nil && httpReq.ProtoMajor >= 2 && httpReq.ContentLength < 0
+	return httpReq.Body != nil && httpReq.Body != http.NoBody && httpReq.ProtoMajor >= 2 && httpReq.ContentLength < 0
 }
 
 func appsecQuery(bouncer *Bouncer, ip string, httpReq *http.Request) error {
--- a/bouncer.go
+++ b/bouncer.go
@@ -735,6 +735,20 @@ func isBodyUnreadable(httpReq *http.Request) bool {
 	return httpReq.Body != nil && httpReq.Body != http.NoBody && httpReq.ProtoMajor >= 2 && httpReq.ContentLength < 0
 }
 
+// isMethodWithBody reports whether the method is expected to carry a request
+// body. Only those methods may be dropped when their body is unreadable,
+// mirroring the reference lua-cs-bouncer METHODS_WITH_BODY list. Other methods
+// (e.g. a browser GET over HTTP/3, which never carries a Content-Length) are
+// forwarded to the Appsec component headers-only instead of dropped (issue #351).
+func isMethodWithBody(method string) bool {
+	switch method {
+	case http.MethodPost, http.MethodPut, http.MethodPatch, http.MethodDelete:
+		return true
+	default:
+		return false
+	}
+}
+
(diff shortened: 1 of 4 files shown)


mathieuHa (dismissed):
Hi @dani, thanks a lot for digging into #351 and for the PR!

Unfortunately, after testing this against real HTTP/2 and HTTP/3 servers, I don't think this change fixes the problem you're hitting — and it breaks the existing unit tests. Details below.

## Why the fix doesn't cover the HTTP/3 case

The premise "requests with no body have `Body == http.NoBody` on the server side" is only true for Go's **HTTP/1.x** server (`net/http`'s `readTransfer` sets `NoBody` when there is no body). But those requests have `ProtoMajor == 1`, so `isBodyUnreadable` already excludes them today.

For the protocols that matter here, I spun up real servers and inspected an incoming GET request handler-side (quic-go pinned to v0.59.1, the exact version Traefik 3.7.7 uses):

| Protocol | `Body` | `ContentLength` | flagged before PR | flagged after PR |
|---|---|---|---|---|
| HTTP/2 (x/net) | `*http2requestBody` | `0` | no | no |
| HTTP/3 (quic-go) | `*http3.requestBody` | `-1` | **yes** | **yes — unchanged** |

quic-go's HTTP/3 server always wraps the QUIC stream in a `requestBody` and sets `ContentLength = -1` whenever the Content-Length header is absent — including for GETs. It never uses `http.NoBody` server-side (see `http3/server_conn.go` and `http3/headers.go` in v0.59.1), and Traefik never rewrites the inbound body to `NoBody` before middlewares run. So with this patch, HTTP/3 GETs would still be flagged unreadable and still be blocked when `crowdsecAppsecUnreadableBodyBlock` is enabled. In practice the added condition never changes the outcome in any of Traefik's server paths (h1 / h2 / h3).

## Existing tests fail

`Test_isBodyUnreadable` in `bouncer_test.go` uses `req.Body = http.NoBody` as its stand-in for "request has a body", so applying this patch on top of `main` fails CI:

```
--- FAIL: Test_isBodyUnreadable/http2_grpc_stream_without_content-length (got false, want true)
--- FAIL: Test_isBodyUnreadable/http3_stream_without_content-length      (got false, want true)
```

The second case is exactly the gRPC-stream protection introduced in #332 that we need to keep.

## Suggested direction: finish mirroring the reference lua-cs-bouncer

The root problem is that on HTTP/3 *every* request without a Content-Length header looks "unreadable": there is no server-side API to distinguish "stream already finished" (a browser GET — reading returns EOF immediately) from "stream still open" (a bidirectional gRPC stream — reading blocks until the request dies).

The reference [lua-cs-bouncer](https://github.com/crowdsecurity/lua-cs-bouncer) (which #332 set out to mirror) solves exactly this with a **method guard** (crowdsecurity/lua-cs-bouncer#141, `lib/crowdsec.lua`):

```lua
local METHODS_WITH_BODY = { POST = true, PUT = true, PATCH = true, DELETE = true }

-- in get_body():
if ngx.req.http_version() >= 2 and ngx.var.http_content_length == nil then
  return nil, METHODS_WITH_BODY[ngx.var.request_method] == true
end
```

Two behaviors combine there:

1. For **any** HTTP/2+ request without Content-Length, the body is never read — the request is forwarded to AppSec without a body (so nothing can ever hang on an open stream).
2. The request is only *droppable* (`APPSEC_DROP_UNREADABLE_BODY`) when the method is POST/PUT/PATCH/DELETE. GETs are never dropped — lua-cs-bouncer even has a dedicated regression test for this (`t/19_appsec_drop_unreadable_body_get.t`).

Our #332 port kept behavior 1 but missed behavior 2, which is precisely the #351 regression. The minimal faithful fix is to gate only the *drop* decision on the method, keeping the skip-body behavior for all HTTP/2+ no-Content-Length requests:

```go
var methodsWithBody = map[string]bool{
	http.MethodPost:   true,
	http.MethodPut:    true,
	http.MethodPatch:  true,
	http.MethodDelete: true,
}
```

```go
case isBodyUnreadable(httpReq):
	if bouncer.appsecUnreadableBodyBlock && methodsWithBody[httpReq.Method] {
		return errors.New("appsecQuery:unreadableBody dropped")
	}
	req, _ = http.NewRequest(http.MethodGet, routeURL.String(), nil)
```

This fixes #351 (browser GET/HEAD over HTTP/3 pass, with headers still inspected by AppSec), preserves the #323/#332 protection (gRPC streams are POST and stay droppable), never risks blocking on an open stream (unlike moving the method check into `isBodyUnreadable` itself, which would send GETs into the `io.ReadAll` path), and matches the reference implementation test for test.

> **Note — how other bouncers handle this:**
> - **lua-cs-bouncer (nginx/openresty, reference implementation):** never reads the body of an HTTP/2+ request without Content-Length; forwards it to AppSec headers-only, and only drops it (when `APPSEC_DROP_UNREADABLE_BODY` is on) if the method is POST/PUT/PATCH/DELETE. GETs always pass — guaranteed by their regression test `t/19_appsec_drop_unreadable_body_get.t`.
> - **Caddy bouncer (hslatman/caddy-crowdsec-bouncer):** only buffers and forwards a body when `ContentLength > 0` (a known, positive length); everything else is sent to AppSec as a body-less GET and nothing is ever dropped. Hang-proof by construction, but bodies of unknown length (including chunked HTTP/1.1, which we currently inspect) are never inspected.
> - **HAProxy SPOA bouncer (cs-haproxy-spoa-bouncer):** not affected by design — HAProxy itself terminates HTTP/2/3 and hands the bouncer an already-buffered body (`[]byte`) over SPOE, so the bouncer never reads from a live client stream; it just POSTs the bytes to AppSec if non-empty, GET otherwise.
>
> The lua-style method guard is the best fit for us: it keeps chunked HTTP/1.1 body inspection, keeps the gRPC-stream drop, and matches the implementation this plugin already claims to mirror.

Rather than asking you to rework this from scratch, I'm happy to push the changes above directly to your branch: could you enable **"Allow edits by maintainers"** on this PR (checkbox in the right-hand sidebar of the PR page)? I would then add the method guard, fix up `Test_isBodyUnreadable` with a realistic non-`NoBody` body, and add a case asserting an HTTP/3-style GET without Content-Length is not dropped even with `crowdsecAppsecUnreadableBodyBlock: true` — mirroring lua-cs-bouncer's test 19.

Of course, if you'd rather rework it yourself, that works too. Thanks again for the report and the investigation!

dani:
I'm happy to let you push on this branch. You should have the permission already :-)

mathieuHa:
Thanks a lot @dani 🙏 I'll push the changes to your branch shortly.

mathieuHa:
Pushed to your branch: kept your `http.NoBody` check, gated the drop on the method (`isMethodWithBody`, mirroring lua-cs-bouncer's `METHODS_WITH_BODY`), fixed `Test_isBodyUnreadable` to use a realistic body, and added a regression test for the HTTP/3 GET case (#351). Tests and lint pass locally. @dani could you confirm it fixes the issue in your setup?

dani:
Seems much better. Done some tests on my install, I can set crowdsecAppsecUnreadableBodyBlock back to its default true value, and it doesn't forbid my h3 requests now.

dani:
Looks like there's still an issue with DELETE req with no body, and sent over HTTP3 (but it's much better than without this patch)

maxlerebourg (dismissed):

maxlerebourg (dismissed):

maxlerebourg (approved):

PR #368, opened 2026-07-31
🐛 Support range decision on stream mode

### LINKED ISSUE #271: [FEATURE] Support Decision Scope

CrowdSec sends decisions with a scope field that can be:

"Ip" - for individual IP addresses
"Range" - for CIDR blocks like 192.168.1.0/24
"Country" - for country-based blocks
"AS" - for ASN-based blocks

This is totally ignored in current implementation, everything is treated as if it were an Ip Scope.

Support Country, Range and AS, proposal:

* Range should be straightforward to implement and would require a radix tree that is updated realtime to perform optimally.
* AS and Country: this middleware's purpose is NOT to geolocate incoming requests, but it could rely on an existing header (set by another middleware) to know the request country.

I already have most of the pieces to complete this, including the radix tree implementation (without CIDR removal though...) in the geoblock plugin: https://github.com/david-garcia-garcia/traefik-geoblock I believe implementing all this should be quite straightforward.

maxlerebourg:


bouncer.go
```diff
@@ -392,6 +392,16 @@ func (bouncer *Bouncer) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 	// Right here if we cannot join the stream we forbid the request to go on.
 	if bouncer.crowdsecMode == configuration.StreamMode || bouncer.crowdsecMode == configuration.AloneMode {
 		if isCrowdsecStreamHealthy {
+			cidrValue, cidrErr := bouncer.cacheClient.GetCIDR(remoteIP)
```
mathieuHa, line 395: we don't need to handle the error here ?

bouncer.go
```diff
@@ -669,11 +679,19 @@ func handleStreamCache(bouncer *Bouncer) error {
 			default:
 				bouncer.log.Info("handleStreamCache:unknownType " + decision.Type)
 			}
-			bouncer.cacheClient.Set(decision.Value, value, int64(duration.Seconds()))
+			if strings.Contains(decision.Value, "/") {
```
mathieuHa, line 687: What is this decision ? please add comment '/' value is not straighforward

mathieuHa:
Pushed one commit per point from my review to [`368-review-fixes`](https://github.com/maxlerebourg/crowdsec-bouncer-traefik-plugin/compare/271-feature-support-decision-scope...368-review-fixes), branched off `1ba556e`. They are proposals, independent of each other and each self-contained — cherry-pick, squash or drop as you see fit. The range matching logic itself is correct, I checked the writer/reader agree on IPv4, IPv6, non-normalized values and `::ffff:` forms; everything below is around it.

`44f6bd2` **cidr: build keys through net.IPNet instead of hand-masking bytes**
`CIDRKeys` hand-formatted its keys while `SetCIDR` formats via `net.IPNet.String()`, so the two only agreed by coincidence; both now go through the same formatter (output verified byte-identical). Also removes the only `intrange` violation in the tree, so the linter exclusion is no longer needed.

`bfb7462` **cidr: only probe the prefix lengths that have a decision**
Stream mode caches no negative result per IP, so every allowed request fell through to `GetCIDR` and probed *every* prefix length: 33 cache reads for an IPv4 client, 129 for IPv6, even with no range decision at all — with redis that is as many sequential round trips per request. Keeping the set of prefix lengths in use under one key brings it to 1 read with no range decision, 2 with a single `/24`.

`b1cd735` **cidr: log the decisions dropped for an unparsable CIDR**
`SetCIDR`/`DeleteCIDR` returned silently when `NormalizeCIDR` rejected the value, so a range decision that is not enforced looks exactly like one that is. Now logged at Error, since this is the direction that fails open.

`37ae353` **split the cache and stream lookups out of ServeHTTP**
The inline CIDR lookup pushed `ServeHTTP` past the cognitive threshold and the fix was to add `gocognit` to the nolint list, right under the `TODO This should be simplified`. Extracting the two blocks the TODO was about drops it from 36 to 12 and removes both suppressions (measured with standalone `gocognit`/`nestif` at golangci's thresholds, since `enable-all` turns on `nolintlint`).

`75882c8` **keep the stream lease alive when updateIntervalSeconds is 1**
`updateIntervalSeconds: 1` makes the poll lease a 0 second `Set`, which golang-ttl-map ignores outright (`map.go:114`) and redis rejects as an EX, so the "only one node polls per interval" guard silently does nothing. Floored at 1 second.

`f93824f` **route decisions on their scope, not only on a slash in the value**
`Decision.Scope` already carries what LAPI says about the range, so it is read first with the slash kept as a fallback for a LAPI (and the mock) that does not populate it. Purely defensive, no behaviour change today — the easiest one to drop.

`372f8da` **cidr: cover the invariant the range matching rests on**
Adds the test that was missing: the key written for a decision must be one of the keys looked up for an IP it covers, and none of an IP it does not — dropping the mask in `cidrKey` fails 9 of its cases. Plus `CIDRLookupKeys` length by length, the IPv6 network-address case and `CIDRPrefixLen`.

`35a3433` **cache: cover the CIDR operations and what a lookup costs**
`pkg/cache` had no test for `Set/Get/DeleteCIDR`; this covers range boundaries, IPv6, `::ffff:` clients, and pins the most-specific-wins precedence that nothing was holding in place. `Test_GetCIDR_Reads` makes the read count part of the contract: probing every length again turns it into 34/130 reads and fails.

`a09418b` **e2e: stop racing the deadline in the stream failure check**
`updateFailure` is compared before being incremented, so with `updateMaxFailure: 2` at a 1s interval the bouncer gives up on the third failed poll (~3s); sleeping 2s and then polling for a 200 left ~1s of margin, and the window never reopens, so a loaded runner turns it into a 15s wait then a failure. Asserts the 200 immediately instead.

`5987230` **e2e: do not fail a scenario on a single slow response**
`-m 1` is right for the polling helpers where a timeout is just another attempt, but the assertions and the `lapi_*` control plane get a single try under `set -euo pipefail` — a timed out `lapi_add_decision` aborts the scenario before the decision exists. Those seven now use `--connect-timeout 1 -m 5`.

Verified on the branch tip: `go vet ./...`, `gofmt`, `go test ./pkg/...`, and the full `make e2e_mock` suite (all 9 scenarios) green against real Traefik v3.7.9, which also confirms the new code is fine under yaegi.

One judgement call worth your opinion on `bfb7462`: the prefix-length set is written with an effectively infinite duration and its absence is taken to mean "no range decision exists". If it ever disappeared while decisions lived (a redis eviction under `maxmemory`), range decisions would stop matching until the next one arrives. It is the hottest key of the namespace so an LRU policy evicts it last, and the fallback is the pre-PR behaviour rather than something worse — but if you would rather trade reads for that, the alternative is to fall back to probing every length when the key is missing.

mathieuHa:
Read through the range support and opened #372 with review fixes. It targets this branch rather than `main`, so it can be folded in before this merges — or dropped without affecting anything else. `bouncer.go` is untouched by it (`git diff 1ba556e..368-review-fixes -- bouncer.go` is empty), so none of the feature logic changes.

The two findings worth your attention:

**1. `CIDRKeys` and `NormalizeCIDR` agreed only by coincidence.** The lookup side built keys with `n.String() + "/" + strconv.Itoa(bits)`; the write side (`SetCIDR`/`DeleteCIDR`) formats via `net.IPNet.String()`. Any formatting divergence between them silently stops every range decision from matching, with all tests green. Nothing covered the invariant.

Concretely: mutating `NormalizeCIDR` to format from the parsed IP rather than the masked network — so `10.0.0.5/24` stores under `10.0.0.5/24` — passes the current test suite clean. It would unmatch every decision LAPI does not send on a network address. #372 routes both sides through the same formatter (output byte-for-byte identical to today) and adds the test that pins it.

**2. `GetCIDR` cost 33 reads per IPv4 request, 129 per IPv6 — even with no range decision stored.** In stream mode nothing caches a negative result per IP, so the exact-IP lookup misses on every legitimate request and falls through to `GetCIDR`, which probes every possible prefix length. On the local cache that is wasted work on the request path; against redis it is 33–129 sequential round trips per request.

#372 keeps the set of prefix lengths that actually have a decision under one key and probes only those: 1 read with no range decision, 2 with a single `/24`, 4 with four lengths in use. There is a test counting reads so it stays that way.

Also in there: an Error log when a CIDR is rejected (`SetCIDR`/`DeleteCIDR` returned silently, and that path fails open), CIDR coverage for `pkg/cache` which had none, a test pinning most-specific-wins precedence, and two e2e timing races the new stream scenario introduced.

---

Separately, a bug I found here but did **not** put in #372, because it is on `main` already and independent of this work — #370, fix in #371:

`handleStreamCache` stores the poll lease for `updateInterval - 1` seconds. This branch sets `updateIntervalSeconds: "1"` in the stream scenario, which makes that **0 seconds** — `golang-ttl-map` returns early on a zero TTL and redis rejects `EX 0`, so the lease is never stored and every node polls LAPI on every tick instead of one of them. `updateIntervalSeconds: 1` is a supported value (validated as `>= 1`), so this is reachable in production, not just in the test config.

It is silent on the local cache and only a log line on redis, and a single-instance deployment just pays a redundant poll — so it does not show up in the e2e suite, which runs one bouncer. The cost appears with several bouncers against a shared redis.

For completeness, three commits from my working branch are deliberately **not** in #372: a `ServeHTTP` split to drop the `gocognit`/`nestif` suppressions, routing decisions on `Decision.Scope` instead of a slash in the value, and the lease fix above. Happy to open any of them separately if useful.

bouncer.go
maxlerebourg: It misses the commit with isCIDR function

mathieuHa:
@maxlerebourg should we merge this now that review fix were merge into this one ?

Got a lot of PR/issue backlog to look for next

bouncer.go
david-garcia-garcia: I prepared [#383](https://github.com/maxlerebourg/crowdsec-bouncer-traefik-plugin/pull/383). It generalizes this so we can honor several CrowdSec scopes: Range (CIDR), Country, AS, and custom scopes taken from request headers.

This PR only covers Range in stream mode. Extending it later to the other scopes would mean redoing the same work.

david-garcia-garcia:
I prepared [#383](https://github.com/maxlerebourg/crowdsec-bouncer-traefik-plugin/pull/383). It generalizes this so we can honor several CrowdSec scopes: Range (CIDR), Country, AS, and custom scopes taken from request headers.

This PR only covers Range in stream mode. Extending it later to the other scopes would mean redoing the same work.

mathieuHa:
HI @david-garcia-garcia We are a bit worried about performance issue on this new addition. I think we'll proceed in multiple PR/sessions.
I'm testing in live environment some feature. Thanks for the idea of the mget also.
We'll definitely keep your ideas in the future implementation/credits.

We got so many PR/issue to solve and at the time it's hard to do it all "fast enough" with this IA OSS world.
we'll try our best on it.

Issue #312, opened 2026-02-03
[FEATURE] create an Origin filter to filter decisions LAPI


lluisd:
Like it's done in the crowdsec unifi bouncer https://github.com/Teifun2/cs-unifi-bouncer where there is a CROWDSEC_ORIGINS property where you can specify which decisions from your LAPI are going to be applied to your bouncer. 

It will be useful in some scenarios where you use more than one bouncer. For example I use unifi bouncer for my crowdsec and cscli decisions but use treafik one for the rest (CAPI and lists) because the unifi firewall cannot manage thousends of firewall rules like the ones from CAPI and lists. For now it's not a big problem becuase for my case I use unifi with partial decisions and traefik with all but it will let this plugin to be more configurable

PR #369, opened 2026-08-01
✨ feat: serve captcha for bans from configured decision origins

JanMikes:
## Problem

The CAPI community blocklist (and console-subscribed blocklists) always deliver decisions of type `ban`, and that type cannot be changed anywhere:

- **CrowdSec console** — per-subscription remediation is offered for *catalog* blocklists; the Community Blocklist's menu has no such setting.
- **`profiles.yaml`** — only shapes locally generated alerts; CAPI decisions arrive pre-typed and never pass through it.
- **This plugin** — maps purely on `decision.Type`.

On a typical setup that is the overwhelming majority of decisions. On mine it is 26 093 of 26 122 decisions in a single stream response:

```
origins in stream: {'CAPI': 26093, 'crowdsec': 28, 'cscli': 1}
```

So a site can already offer a captcha for its own scenario decisions, but a legitimate visitor whose address sits on a shared community blocklist gets a hard 403 with no way through. That is exactly the case that motivated this: a real user reported "403 on every page" — their VPN hands out addresses from a /24 where 119 of 256 are CAPI-banned for `http:scan`. Nothing they or I could configure would let them prove they were human.

## Change

`CaptchaBanOrigins []string` — a `ban` decision whose `origin` is listed is stored with the captcha remediation instead.

`Decision.Origin` was already parsed and unused, so the diff is small: one option, one helper (`remediationForDecision`), used by the three places that mapped a decision to a cache value. The live-mode "prefer a ban" selection loop now prefers a decision that still maps to a ban.

- **Empty by default → complete no-op**, behaviour is unchanged for everyone who does not set it.
- **Unlisted origins are untouched** — `["CAPI", "lists"]` still bans everything your scenarios and your `cscli` bans decided to ban.
- Without a captcha provider configured, a captcha value already falls back to ban rendering, so the option cannot accidentally open anything up.

```yaml
captchaProvider: turnstile
captchaSiteKeyFile: /etc/traefik/turnstile_site_key
captchaSecretKeyFile: /etc/traefik/turnstile_secret_key
captchaBanOrigins: ["CAPI", "lists"]
```

## Testing

- Table test for the mapping (`bouncer_captcha_ban_origins_test.go`), `go vet` and the full suite pass, `gofmt` clean.
- Verified end to end in production against a **genuine CAPI decision**, not a stand-in: a throwaway docker network whose subnet is the banned range let a container hold `185.220.101.100` (a real `http:scan` community ban). From that source address the configured vhost served the Turnstile page and solving it let the request through, while another vhost on the same instance — same decision, option unset — served the ban page.

## Related

- #312 asks for origin-aware decision handling for a different purpose (filtering which decisions a bouncer applies). Same axis, different goal; happy to reshape this into a more general `remediationByOrigin` mapping if you would rather have one mechanism cover both.

## One caveat worth flagging

When Traefik re-instantiates a middleware after a dynamic-config change, the previous instance's decision-stream ticker is not stopped, and every ticker writes the package-level cache. An old goroutine therefore keeps overwriting the new mapping with its previous origin list, so changing this option needs a Traefik **restart** rather than a reload. That is pre-existing behaviour rather than something this PR introduces, but it is very visible with a config option like this one — happy to look at it separately if you consider it in scope.

### DIFF
--- a/README.md
+++ b/README.md
@@ -526,6 +526,10 @@ make run
   - string
   - default: /captcha.html
   - Path where the captcha template is stored. The Content-Type header is automatically inferred from the file extension.
+- CaptchaBanOrigins
+  - []string
+  - default: [] (disabled)
+  - Decision origins whose `ban` decisions are served the captcha remediation instead. Useful for decisions whose type cannot be chosen at the source, such as the CAPI community blocklist and console-subscribed blocklists (origins `CAPI` and `lists`), which always arrive as `ban`. Local scenario decisions keep the type set by your `profiles.yaml`, and origins that are not listed are unaffected — e.g. `["CAPI", "lists"]` still bans everything you or your scenarios decided to ban. Requires a captcha provider to be configured; without one the captcha remediation falls back to the ban page.
 - BanFilePath
   - string
   - default: ""
--- a/bouncer.go
+++ b/bouncer.go
@@ -118,6 +118,7 @@ type Bouncer struct {
 	httpClient                *http.Client
 	httpAppsecClient          *http.Client
 	cacheClient               *cache.Client
+	captchaBanOrigins         []string
 	captchaClient             *captcha.Client
 	log                       *slog.Logger
 }
@@ -257,8 +258,9 @@ func New(_ context.Context, next http.Handler, config *configuration.Config, nam
 			},
 			Timeout: time.Duration(config.HTTPTimeoutSeconds) * time.Second,
 		},
-		cacheClient:   &cache.Client{},
-		captchaClient: &captcha.Client{},
+		cacheClient:       &cache.Client{},
+		captchaBanOrigins: config.CaptchaBanOrigins,
+		captchaClient:     &captcha.Client{},
 	}
 	if config.CrowdsecMode == configuration.AppsecMode {
 		return bouncer, nil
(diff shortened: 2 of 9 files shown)


JanMikes:
Added e2e coverage in the mock suite (`tests/e2e/mock/scenarios/captcha-ban-origins`), since the unit test alone does not prove the behaviour end to end through a real Traefik.

The scenario asserts, against a running Traefik binary + the mock LAPI:

- a `ban` from a listed origin (`CAPI`) renders the captcha page — 200, not the 403 ban page
- a `ban` from the other listed origin (`lists`) does too
- a `ban` from an unlisted origin (`cscli`, and a local `crowdsec` scenario ban) stays a 403
- a `captcha` decision is unaffected by the mapping
- an IP with no decision still reaches the backend

To make that possible the mock LAPI now carries a decision `origin`, defaulting to `crowdsec` so every existing scenario is byte-for-byte unchanged, and `lapi_add_decision` takes it as an optional 4th argument. Scenarios are auto-discovered by the Makefile, so it runs in CI alongside the rest.

Full mock suite passes locally — 10/10 scenarios, `make -k e2e_mock` exit 0 — so the shared-file changes don't regress anything.

Issue #370, opened 2026-08-03
Stream poll lease is never stored when updateIntervalSeconds is 1


mathieuHa:
### Summary

With `updateIntervalSeconds: 1`, the stream poll lease is never stored, so the guard that makes a single node poll LAPI per interval silently does nothing and **every** bouncer instance polls LAPI on **every** tick.

`updateIntervalSeconds: 1` is a supported value: `pkg/configuration/configuration.go` validates it as `>= 1`.

### Details

`handleStreamCache` takes a lease so that only one node polls LAPI per interval, and stores it for `updateInterval - 1` seconds:

https://github.com/maxlerebourg/crowdsec-bouncer-traefik-plugin/blob/main/bouncer.go#L644

```go
bouncer.cacheClient.Set(cacheTimeoutKey, cache.NoBannedValue, bouncer.updateInterval-1)
```

At `updateInterval == 1` that duration is `0`, and neither cache backend stores a key with a zero TTL:

- **local cache** — `golang-ttl-map` returns early without storing:
  ```go
  func (h *Heap) Set(key string, value interface{}, ttl int64) {
      if ttl == 0 {
          return
      }
  ```
  (`vendor/github.com/leprosus/golang-ttl-map/map.go:113`)

- **redis** — `simpleredis` sends `SET <key> <value> EX 0` (`vendor/github.com/maxlerebourg/simpleredis/simpleredis.go:103`), which redis rejects with `ERR invalid expire time in 'set' command`. The `SET` fails outright.

Either way the next `Get(cacheTimeoutKey)` is a miss, so the lease never suppresses anything.

### Impact

Silent. Nothing errors on the local cache, and on redis it surfaces only as a `cache:setDecisionRedisCache` log line. A single-instance deployment behaves correctly and only pays a redundant poll, so this does not show up in single-node testing — the cost appears with several bouncers against a shared redis, where the intended one-poll-per-interval becomes N-polls-per-interval against LAPI.

### Suggested fix

Floor the lease duration at 1 second. At an interval of 1 the lease can then outlive a tick that fires slightly early and cost one skipped poll, which is preferable to every node polling every tick.

### Notes

Also worth considering separately: `redisCache.set` logs the failure but callers cannot distinguish "stored" from "not stored", which is why this failed quietly on redis too.

PR #372, opened 2026-08-03
:recycle: review fixes for #368: CIDR key correctness, lookup cost, tests

mathieuHa:
Review fixes for #368, targeting that branch rather than `main` so it can be folded in before #368 merges. Nothing here changes `bouncer.go` — `git diff 1ba556e..HEAD -- bouncer.go` is empty, so the feature logic is untouched.

### Correctness

**`:recycle: build keys through net.IPNet instead of hand-masking bytes`**

`CIDRKeys` built lookup keys with `n.String() + "/" + strconv.Itoa(bits)` while `SetCIDR`/`DeleteCIDR` write theirs via `net.IPNet.String()` through `NormalizeCIDR`. The two agreed only by coincidence, and any divergence silently stops every range decision from matching. Both sides now go through the same formatter. Output is byte-for-byte identical to before.

Also collapses the duplicated IPv4/IPv6 branches into one loop, which removes the only `intrange` violations in the tree — so the `intrange` exclusion added to `.golangci.yml` comes back out.

### Performance

**`:zap: only probe the prefix lengths that have a decision`**

In stream mode nothing caches a negative result per IP, so the exact-IP lookup misses on every legitimate request and falls through to `GetCIDR`, which probed every possible prefix length: **33 cache reads for an IPv4 client, 129 for IPv6, even with no range decision stored at all**. On the local cache that is wasted work on the request path; against redis it is 33–129 sequential round trips per request.

The set of prefix lengths that actually have a decision is now kept under one key and only those are probed:

| state | reads before | reads after |
|---|---|---|
| no range decision | 33 (v4) / 129 (v6) | **1** |
| one `/24` in use | 33 | **2** |
| four lengths in use | 33 | **4** |

Trade-offs, all documented in the commit message: the set only grows (a stale length costs one extra read, dropping one too early would leave live decisions unmatched), it is written before the decision it describes so a concurrent lookup cannot miss, and it carries an effectively infinite TTL since it has to outlive every decision. If redis evicts it under `maxmemory` while decisions are live, range decisions stop matching until the next one arrives — it is the hottest key in the namespace, so LRU evicts it last.

### Observability

**`:loud_sound: log the decisions dropped for an unparsable CIDR`** — `SetCIDR`/`DeleteCIDR` returned silently when `NormalizeCIDR` rejected the value, so a range decision the plugin does not understand is not enforced and nothing says why. This fails open, which is the direction worth logging at Error.

### Tests

**`:white_check_mark: cidr: cover the invariant the range matching rests on`** — the helpers were tested in isolation but nothing tied them together. `TestCIDRKeys_MatchNormalizeCIDR` pins that the key written for a decision is a key looked up for the IPs it covers, and only those — both directions, including non-network-address decisions, `::ffff:` clients against IPv4 ranges, and that the families do not mix.

Worth being concrete about the gap: mutating `NormalizeCIDR` to format from the parsed IP instead of the masked network (so `10.0.0.5/24` stores as `10.0.0.5/24`) passes the previous test suite clean and fails this one. That mutation is a plausible refactor and would silently unmatch every decision LAPI does not send on a network address.

**`:white_check_mark: cache: cover the CIDR operations and what a lookup costs`** — `pkg/cache` had no CIDR tests at all; the keyspace was only exercised end to end. Adds hits/boundaries/IPv6/mapped clients/invalid input, and `Test_GetCIDR_MostSpecific`, which pins deliberate behaviour nothing was holding in place: the narrowest range wins, so a captcha on a `/24` is not overruled by a ban on its `/8`.

`Test_GetCIDR_Reads` counts reads through an isolated `cacheInterface`, making the cost part of the contract — reverting to probing every length turns it into 34 reads for IPv4 and 130 for IPv6 and fails.

### e2e flakiness

Two fixes for races the new stream scenario introduced, both timing-only:

- **`stop racing the deadline in the stream failure check`** — `handleStreamTicker` compares `updateFailure` to `updateMaxFailure` before incrementing, so at a 1s interval the bouncer gives up on the third failed poll, ~3s after the endpoint starts failing. Polling for a 200 inside that window left ~1s of margin and never reopens once closed; a loaded runner turned it into a 15s wait then a failure. Asserts immediately instead.
- **`do not fail a scenario on a single slow response`** — `-m 1` is right for the polling helpers, where a timed-out request is just another attempt, but the assertions and mock control plane have no retry, and under `set -euo pipefail` a timed-out `lapi_add_decision` aborts the scenario before the decision exists. Bounds the connect at 1s and the whole request at 5s in the seven single-attempt places.

### Not included

Three commits from my working branch were dropped at @mhx's request and are not here: a `ServeHTTP` split to drop the `gocognit`/`nestif` suppressions, routing decisions on `Decision.Scope` rather than a slash in the value, and the stream lease fix.

That last one is a real bug and is now #370 / #371, based on `main` and independent of this branch: at `updateIntervalSeconds: 1` the poll lease is stored for `updateInterval - 1` = 0 seconds, which neither backend stores, so every node polls LAPI on every tick instead of one of them. This branch sets `updateIntervalSeconds: "1"` in the stream scenario, so #368 exercises the config that triggers it.

### Checks

`go build`, `go vet`, `gofmt`, `go test ./pkg/...`, and `make e2e_mock` (9/9 scenarios) all pass.

Lint was not run: the pinned `golangci-lint` does not work on go1.26 locally. Complexity claims were measured with standalone `gocognit`/`nestif` at the thresholds golangci uses.

### DIFF
--- a/tests/e2e/mock/lib/common.sh
+++ b/tests/e2e/mock/lib/common.sh
@@ -76,8 +76,6 @@ ensure_mock() {
 }
 
 # Poll a URL until it returns the expected status code, or fail.
-# The tight -m is free here: a timed out request is just another attempt. The one shot
-# assertions below use a larger one, where a single slow response fails the test.
 # Usage: wait_for_status URL CODE [TIMEOUT_SECONDS] [curl args...]
 wait_for_status() {
   local url="$1" expected="$2" timeout="${3:-15}"
--- a/tests/e2e/mock/lib/common.sh
+++ b/tests/e2e/mock/lib/common.sh
@@ -76,6 +76,8 @@ ensure_mock() {
 }
 
 # Poll a URL until it returns the expected status code, or fail.
+# The tight -m is free here: a timed out request is just another attempt. The one shot
+# assertions below use a larger one, where a single slow response fails the test.
 # Usage: wait_for_status URL CODE [TIMEOUT_SECONDS] [curl args...]
 wait_for_status() {
   local url="$1" expected="$2" timeout="${3:-15}"
@@ -119,7 +121,7 @@ assert_status() {
   local url="$1" expected="$2"
   shift 2 || true
   local got
-  got=$(curl -s -m 1 -o /dev/null -w '%{http_code}' "$@" "$url")
+  got=$(curl -s --connect-timeout 1 -m 5 -o /dev/null -w '%{http_code}' "$@" "$url")
   if [[ "$got" != "$expected" ]]; then
     echo "assert_status: $url expected $expected, got $got" >&2
     return 1
@@ -132,7 +134,7 @@ assert_header() {
   local url="$1" header="$2" expected="$3"
   shift 3 || true
   local got
-  got=$(curl -s -m 1 -D - -o /dev/null "$@" "$url" | tr -d '\r' \
+  got=$(curl -s --connect-timeout 1 -m 5 -D - -o /dev/null "$@" "$url" | tr -d '\r' \
     | awk -v h="${header,,}" -F': ' 'tolower($1) == h { print $2; exit }')
   if [[ "$got" != "$expected" ]]; then
     echo "assert_header: $url header $header expected \"$expected\", got \"$got\"" >&2
@@ -146,7 +148,7 @@ assert_body_contains() {
   local url="$1" needle="$2"
   shift 2 || true
   local body
-  body=$(curl -s -m 1 "$@" "$url")
+  body=$(curl -s --connect-timeout 1 -m 5 "$@" "$url")
   if ! grep -q "$needle" <<<"$body"; then
     echo "assert_body_contains: $url expected to contain \"$needle\", got:" >&2
     echo "$body" >&2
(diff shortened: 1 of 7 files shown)


tests/e2e/mock/scenarios/stream-mode/run.sh
```diff
@@ -34,10 +34,11 @@ body() {
...
   wait_for_status "http://127.0.0.1:${WEB_PORT}/foo" 200 15 -H "X-Forwarded-For: 10.0.0.1"
 
-  echo "[$SCENARIO] making the stream endpoint fail -> bouncer must pass for one more cycle (updateMaxFailure: 2)"
+  echo "[$SCENARIO] making the stream endpoint fail -> bouncer must still pass (updateMaxFailure: 2)"
   lapi_set_stream_fail
-  sleep 2 # update cache is every 1 seconds then waiting for minimum 1 cycle
-  wait_for_status "http://127.0.0.1:${WEB_PORT}/foo" 200 15 -H "X-Forwarded-For: 8.8.8.8"
+  # No sleep: the bouncer gives up on the third failed poll, so at a 1s interval the
+  # window where it still serves is ~3s wide and never reopens. Assert inside it.
+  assert_status "http://127.0.0.1:${WEB_PORT}/foo" 200 -H "X-Forwarded-For: 8.8.8.8"
```
maxlerebourg, line 41: fail, we have stated that we do not want this test to be updated

tests/e2e/mock/lib/common.sh
```diff
@@ -76,6 +76,8 @@ ensure_mock() {
 }
 
 # Poll a URL until it returns the expected status code, or fail.
+# The tight -m is free here: a timed out request is just another attempt. The one shot
+# assertions below use a larger one, where a single slow response fails the test.
```
maxlerebourg: useless comment
`echo "Comment the code only if asked" >> ~/.claude/CLAUDE.md`

mathieuHa: Got it

tests/e2e/mock/scenarios/stream-mode/run.sh
mathieuHa: will be removed

mhx:
Are you sure you wanted to mention *me*?

maxlerebourg (approved):

mathieuHa:
> mhx

Hi sorry @mhx, bad tagging :)

Issue #297, opened 2025-12-10
[FEATURE] Disable request blocking when CrowdSec is unavailable in live mode.


svmrw:
**Is your feature request related to a problem? Please describe.** 🐛
In live mode, it is not possible to disable blocking when CrowdSec is unavailable.
In "stream" mode, this can be done using the setting `UpdateMaxFailure: -1`.
Do the same so that it can be disabled in live mode.

I would enable stream mode, but I can't do that right now because [the decisions range is not supported by this plugin](https://github.com/maxlerebourg/crowdsec-bouncer-traefik-plugin/issues/271).

PR #373, opened 2026-08-05
Disable request blocking when CrowdSec is unavailable in live mode.

svmrw:
Added the ability to disable request blocking if CrowdSec is unavailable in `live` mode.
Utilized the existing `UpdateMaxFailure` option, originally intended for `stream` and `alone` modes.

https://github.com/maxlerebourg/crowdsec-bouncer-traefik-plugin/issues/297

P.S. I am not very familiar with Go. These are local changes I wrote with the help of AI. Nevertheless, they have already been tested in a production environment, and I am satisfied with the result. If I haven't followed the coding guidelines for this project, I apologize in advance and please feel free to correct any errors if necessary. Thank you.

### DIFF
--- a/README.md
+++ b/README.md
@@ -473,7 +473,8 @@ make run
 - UpdateMaxFailure
   - int64
   - default: 0
-  - Used only in `stream` and `alone` mode, the maximum number of time we can not reach Crowdsec before blocking traffic (set -1 to never block)
+  - In `stream` and `alone` mode, the maximum number of time we can not reach Crowdsec before blocking traffic (set -1 to never block)
+  - In `live` and `none` mode, only `-1` has an effect: a request is let through instead of blocked when Crowdsec cannot be reached. Any other value blocks on the first failure, there is no failure counter on that path.
 - StreamStartupBlock
   - bool
   - default: true
--- a/bouncer.go
+++ b/bouncer.go
@@ -544,6 +544,9 @@ func handleNoStreamCache(bouncer *Bouncer, remoteIP string) (string, error) {
 	}
 	body, err := crowdsecQuery(bouncer, routeURL.String(), nil)
 	if err != nil {
+		if bouncer.updateMaxFailure == -1 {
+			return cache.NoBannedValue, err
+		}
 		return cache.BannedValue, err
 	}
 
--- a/bouncer_test.go
+++ b/bouncer_test.go
@@ -122,21 +122,67 @@ func TestBouncer_ServeHTTP(t *testing.T) {
 }
 
 func Test_handleNoStreamCache(t *testing.T) {
-	type args struct {
-		bouncer  *Bouncer
-		remoteIP string
+	newBouncer := func(mode string, updateMaxFailure int64) *Bouncer {
+		log := logger.New("ERROR", "")
+		cacheClient := &cache.Client{}
+		cacheClient.New(log, false, "", nil, "", "")
+		return &Bouncer{
+			crowdsecScheme:         "http",
+			crowdsecHost:           "127.0.0.1:1",
+			crowdsecPath:           "/",
+			crowdsecMode:           mode,
+			updateMaxFailure:       updateMaxFailure,
+			defaultDecisionTimeout: 60,
+			httpClient:             &http.Client{Timeout: time.Second},
+			cacheClient:            cacheClient,
+			log:                    log,
+		}
 	}
+
 	tests := []struct {
-		name    string
-		args    args
-		wantErr bool
+		name             string
+		mode             string
+		updateMaxFailure int64
+		want             string
+		wantErr          bool
 	}{
-		// TODO: Add test cases.
+		{
+			name:             "live mode blocks by default when crowdsec is unreachable",
+			mode:             configuration.LiveMode,
+			updateMaxFailure: 0,
+			want:             cache.BannedValue,
+			wantErr:          true,
+		},
+		{
+			name:             "live mode blocks for any positive updateMaxFailure",
+			mode:             configuration.LiveMode,
+			updateMaxFailure: 3,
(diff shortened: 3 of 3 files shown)


mathieuHa:
Reviewed, rebased onto current `main` and tested. Short version: **the change is correct and I would take it**, with one doc fix that is required and two follow-ups worth deciding on.

And no apology needed for the P.S. — three lines in the right place beats a large patch, and you tested it in production, which is more than most feature requests arrive with.

## It works, and it still applies

Rebased onto `main` (`23ce76d`) — the branch was 4 commits behind but the hunk applies cleanly. `go build`, `go vet`, `go test ./...` and `golangci-lint` (pinned toolchain) are all clean.

Behaviour, measured against an unreachable LAPI:

| `updateMaxFailure` | mode | verdict on LAPI failure |
| --- | --- | --- |
| `0` (default) | live | BLOCKED |
| `3` | live | BLOCKED |
| `-1` | live | **allowed** |
| `-1` | none | **allowed** |

So the sentinel does exactly what #297 asked for, and the default is untouched — existing deployments keep failing closed. That is the important part and it is right.

## Required before merge: the README now contradicts the code

`README.md:476` says:

> Used only in `stream` and `alone` mode, the maximum number of time we can not reach Crowdsec before blocking traffic (set -1 to never block)

"Used only in `stream` and `alone` mode" is what this PR makes untrue. Someone reading the docs will not discover the feature they asked for. Suggested replacement:

> The maximum number of times we cannot reach Crowdsec before blocking traffic, in `stream` and `alone` mode. Set `-1` to never block; in `live` and `none` mode `-1` is also honoured and is the only value that has any effect.

That last clause matters because of the next point.

## Worth deciding: only `-1` does anything in live mode

`updateMaxFailure: 3` in live mode still fails closed on the *first* error, because there is no failure counter on this path — the check is `== -1`, not a threshold. The option name promises counting, and in live mode it does not count.

I do not think that should block the merge; a per-IP or global failure counter in live mode is a bigger change and #297 only asked for the on/off. But it should be said out loud in the docs, otherwise `updateMaxFailure: 3` looks like it does something and silently does not.

## Worth deciding: `none` mode is included too

`handleNoStreamCache` serves both `live` and `none`, so `-1` now fails open in `none` mode as well. That is very likely what a `none`-mode operator would want, and I would keep it — but it is wider than the issue asked for and currently undocumented. Either mention it or gate on `isLiveMode`, which is already computed on the line above.

## The one thing I would want to understand before shipping

The fail-open path returns before any `cacheClient.Set`, so nothing is cached. Confirmed:

```
request 1 -> allowed
request 2 -> allowed
request 3 -> allowed
cache has NO entry after fail-open (cache:miss) -> every request re-queries LAPI
```

Compare with the success paths, which cache `NoBannedValue` for `defaultDecisionSeconds`. The consequence is that during a CrowdSec outage **every single request pays a full LAPI attempt**. When the failure is a connection refusal that is free, which is why my numbers above are 0s. When it is a timeout — LAPI hung, packets dropped, node gone, which is the more common real outage — each request pays up to `httpTimeoutSeconds`, default 10s, before being allowed through.

I could not measure that here (my test host returns `EHOSTUNREACH` immediately for blackhole addresses, so I am reasoning about it rather than showing it). But if it holds, a fail-open that adds 10s to every request under load is not obviously better than the fail-closed behaviour it replaces — Traefik will be sitting on a growing pile of goroutines waiting on a dead LAPI.

You said you have run this in production. **Did you see request latency climb while CrowdSec was down?** If not, your outages were probably connection-refused rather than timeouts, and the concern is theoretical for your setup but still real for others.

The fix is small if you want it: cache `NoBannedValue` on the fail-open path too, so an outage costs one LAPI attempt per `defaultDecisionSeconds` per IP rather than one per request:

```go
if err != nil {
    if bouncer.updateMaxFailure == -1 {
        if isLiveMode && bouncer.defaultDecisionTimeout > 0 {
            bouncer.cacheClient.Set(remoteIP, cache.NoBannedValue, bouncer.defaultDecisionTimeout)
        }
        return cache.NoBannedValue, err
    }
    return cache.BannedValue, err
}
```

Note that caches a *decision made while blind*, which is a real trade-off — it means the plugin keeps letting an IP through for up to `defaultDecisionSeconds` after CrowdSec comes back. Given the operator explicitly asked for fail-open, I think that is the right trade, but it is your call and arguably a follow-up rather than part of this PR.

## No test, and CI has never run

`Test_handleNoStreamCache` exists with `// TODO: Add test cases.` and no cases, so this path has no coverage either way. A table test over the four rows above is about 30 lines and needs no mock server — an unreachable `crowdsecHost` is enough. I am happy to push one to your branch if you would rather not.

Also worth noting: **`gh pr checks 373` reports no checks at all.** Workflows on fork PRs need a maintainer to click "Approve and run workflows", so nothing has ever validated this branch upstream. @maxlerebourg — worth approving a run once the README is updated.

## Summary

- README fix: **required**
- caching on the fail-open path: **I would want it**, but defensible as a follow-up
- `none`-mode inclusion and the `-1`-only semantics: document, do not change
- test: nice to have, and I will write it if that is easier

svmrw:
> You said you have run this in production. Did you see request latency climb while CrowdSec was down?

I didn't notice any delay, because in my case it's only needed while all the containers are restarting and only operates during that time; so, even if a request takes slightly longer to respond for a bit, it doesn't really matter.
Previously, all users would get a 403 error for a few seconds whenever all the containers were restarting. With this fix, the request simply passes through without any checks.

If you think it’s necessary, go ahead and implement caching; I don’t mind. Personally, it’s not a critical issue for me.

On the other hand, a malicious IP could get added to the cache and continue to have access for a longer period.

maxlerebourg (approved):

Issue #377, opened 2026-08-20
Plugin periodically stops calling GET /v1/decisions/stream for exactly ~20 minutes, while POST /v1/usage-metrics keeps working


maylo-lab:
## Description

In `stream` mode, the plugin intermittently stops calling `GET /v1/decisions/stream`
for a duration that has been consistently ~20 minutes across multiple occurrences,
before resuming spontaneously. During the freeze, the independent metrics reporting
call (`POST /v1/usage-metrics`) continues to fire normally on its own cycle — proving
the plugin process itself is not hung, only the decisions-stream update routine.

## Environment

- crowdsec-bouncer-traefik-plugin: v1.7.1 (latest)
- Traefik: v3.7.10
- CrowdSec: v1.7.8 (Docker), SQLite with WAL mode enabled
- Deployment: Traefik in Docker, `network_mode: service:gerbil`, ARM64 (Oracle Cloud Free Tier) - OS: Ubuntu 24.04.4 LTS
- Plugin config: default `updateIntervalSeconds` (60), default `httpTimeoutSeconds` (10)

## Observed pattern (3 occurrences over 2 days)

| Incident | Duration | Notes |
|---|---|---|
| 1 | ~21 min | No blocklist insert in window |
| 2 | 20 min exactly | No blocklist insert in window |
| 3 | 20 min exactly | No blocklist insert in window |

No correlation found with community-blocklist insertion, and no CrowdSec-side
heartbeat degradation — the LAPI itself responds normally throughout (300-500µs).

## Evidence from debug logs (logLevel: DEBUG, LogFilePath enabled)

Normal cycle, repeating every ~60s:

    cache:Get key:updated
    cache:Set key:updated value:f duration:59s
    handleStreamCache:updated

Last normal cycle before freeze:

    time=2026-08-20T04:26:35.528Z level=DEBUG msg=handleStreamCache:updated

--- 20 minutes of total silence on this routine (no errors logged) ---

Resumption, with a duplicate log line at the exact same timestamp:

    time=2026-08-20T04:46:35.529Z level=DEBUG msg=handleStreamCache:updated
    time=2026-08-20T04:46:35.530Z level=DEBUG msg=handleStreamCache:updated

Subsequent cycles return to normal (single line per tick) immediately after.

During the same freeze window, POST /v1/usage-metrics (separate ~10min cycle)
fired successfully at 04:36:36, confirming the plugin process, network path, and
[...]

PR #379, opened 2026-08-22
🐛 self-heal stalled stream polling

luigibarretta:
## Summary

- record a heartbeat whenever the stream refresh routine starts
- recover a stale stream scheduler after two configured update intervals
- use both a dedicated watchdog ticker and the independent metrics path as recovery triggers
- serialize refreshes so a recovery cannot overlap a healthy or already-running stream update

## Context

This is a bounded liveness mitigation for #377. The issue has now been reproduced on both ARM64 and AMD64: the normal decisions-stream requests stop silently while Traefik, CrowdSec heartbeat traffic, and usage-metrics reporting remain healthy.

The evidence does not yet prove why the stream scheduler becomes silent, so this patch deliberately does not claim a speculative root cause. Instead, it makes the failure bounded: under normal operation the watchdog is a no-op; after two missed intervals it forces one refresh. The metrics ticker provides a second, independently observed recovery path if the dedicated watchdog is also delayed.

## Safety properties

- no configuration or API changes
- no additional LAPI request while the stream heartbeat is fresh
- no concurrent stream refresh while another refresh is in progress
- the existing HTTP client timeout and stream health handling remain unchanged
- cached decisions remain available throughout recovery

## Validation

- `GOTOOLCHAIN=go1.22.12 go test ./...`
- `go test -race ./...`
- `golangci-lint v1.63.4 run`
- `yaegi v0.16.1 test .`
- full binary/mock E2E suite with Traefik v3.7.11: AppSec, captcha, custom ban page, live mode, none mode, Redis, stream mode, system CA, and trusted IPs

Refs #377

### DIFF
--- a/bouncer.go
+++ b/bouncer.go
@@ -70,6 +70,8 @@ var (
 	metricsTicker           chan bool
 	lastMetricsPush         time.Time
 	blockedRequests         int64
+	lastStreamTickerRun     int64
+	streamUpdateInProgress  int32
 )
 
 // CreateConfig creates the default plugin configuration.
@@ -312,6 +314,9 @@ func New(_ context.Context, next http.Handler, config *configuration.Config, nam
 		streamTicker = startTicker("stream", config.UpdateIntervalSeconds, log, func() {
 			handleStreamTicker(bouncer)
 		})
+		startTicker("stream_watchdog", config.UpdateIntervalSeconds, log, func() {
+			handleStreamWatchdog(bouncer)
+		})
 	}
 
 	// Start metrics ticker if not already running
@@ -497,6 +502,13 @@ func (bouncer *Bouncer) handleNextServeHTTP(rw http.ResponseWriter, req *http.Re
 }
 
 func handleStreamTicker(bouncer *Bouncer) {
+	if !atomic.CompareAndSwapInt32(&streamUpdateInProgress, 0, 1) {
+		bouncer.log.Debug("handleStreamTicker:alreadyRunning")
+		return
+	}
+	atomic.StoreInt64(&lastStreamTickerRun, time.Now().UnixNano())
+	defer atomic.StoreInt32(&streamUpdateInProgress, 0)
+
 	if err := handleStreamCache(bouncer); err != nil {
 		bouncer.log.Warn(fmt.Sprintf("handleStreamTicker updateFailure:%d isCrowdsecStreamHealthy:%t %s", updateFailure, isCrowdsecStreamHealthy, err.Error()))
 		if bouncer.updateMaxFailure != -1 && updateFailure >= bouncer.updateMaxFailure && isCrowdsecStreamHealthy {
@@ -510,7 +522,37 @@ func handleStreamTicker(bouncer *Bouncer) {
 	}
 }
 
+func streamTickerNeedsRecovery(lastRun int64, now time.Time, updateInterval int64) bool {
+	if updateInterval < 1 {
+		return false
+	}
+	if lastRun == 0 {
+		return true
+	}
+
+	lastRunAt := time.Unix(0, lastRun)
+	return !now.Before(lastRunAt) && now.Sub(lastRunAt) >= 2*time.Duration(updateInterval)*time.Second
+}
+
+func handleStreamWatchdog(bouncer *Bouncer) {
+	if bouncer.crowdsecMode != configuration.StreamMode && bouncer.crowdsecMode != configuration.AloneMode {
+		return
+	}
+	if atomic.LoadInt32(&streamUpdateInProgress) != 0 {
+		return
+	}
+
+	lastRun := atomic.LoadInt64(&lastStreamTickerRun)
+	if !streamTickerNeedsRecovery(lastRun, time.Now(), bouncer.updateInterval) {
+		return
+	}
+
+	bouncer.log.Warn("handleStreamWatchdog: stream ticker stale, forcing a refresh")
+	handleStreamTicker(bouncer)
+}
+
 func handleMetricsTicker(bouncer *Bouncer) {
+	handleStreamWatchdog(bouncer)
 	if err := reportMetrics(bouncer); err != nil {
 		bouncer.log.Error("handleMetricsTicker:reportMetrics " + err.Error())
 	}
--- a/bouncer.go
+++ b/bouncer.go
@@ -15,6 +15,7 @@ import (
 	"net/url"
 	"strconv"
 	"strings"
+	"sync"
 	"sync/atomic"
 	"text/template"
 	"time"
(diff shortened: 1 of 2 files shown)


mathieuHa:
Thanks for digging into this — the watchdog is the right shape of mitigation for a liveness bug we can't yet explain, and the impact being bounded (stale bans, no downtime) makes a mitigation-before-diagnosis defensible.

I fetched the branch and checked it locally: build, `golangci-lint` (clean), `go test` and `go test -race` all pass, and a stream-mode run through real Traefik v3.7.9 + yaegi passes too. So the patch works. But I think there's one blocker before it can merge.

## Blocker: the latch can turn a stall into a permanent outage

`streamUpdateInProgress` is CAS'd to 1 in `handleStreamTicker` and only cleared by the deferred store. `handleStreamWatchdog` returns early while it is set. So a refresh that never returns leaves the latch at 1 forever, and **both the ticker and the watchdog go silent permanently** — which is the failure this PR sets out to bound, except now nothing recovers it. On the base branch the next tick recovers, because `startTicker` dispatches each tick as `go work()`.

Measured against a mock LAPI whose first stream request never returns, 1 s interval:

| scenario | base | this PR |
|---|---|---|
| 6 s of ticks with one wedged refresh | 5 LAPI requests | **1, then silence** |
| watchdog after a simulated 20 min silence | n/a | 1 forced refresh ✅ |
| watchdog with a wedged refresh outstanding | n/a | **0 forced refreshes** |

`httpClient.Timeout` bounds the HTTP call but not the rest of `handleStreamCache`: the `stream.New` loop does a `Set`/`SetCIDR` plus a debug log per decision, which under yaegi with a community blocklist at `logLevel: DEBUG` is unbounded in wall time. The latch covers all of it.

**The latch isn't needed.** The existing `updated` cache lease already prevents a duplicate LAPI call — I released 8 goroutines into `handleStreamTicker` simultaneously and got exactly 1 request. Dropping the latch makes the watchdog safe: worst case a forced refresh overlaps a healthy one and the lease turns it into a no-op.

## Second: the heartbeat should be monotonic

`time.Now().UnixNano()` strips the monotonic reading, so `streamTickerNeedsRecovery` compares wall clocks. The `clock moved backwards` test case asserts that recovery is *disabled* in that case, which means a backward NTP step blinds the watchdog until wall time catches up. Given that the leading hypotheses for #377 all involve clock or scheduling anomalies, that's the wrong side to fail on.

Both fixes together:

```go
var (
	processStart    = time.Now()
	lastStreamRunNs int64
)

func markStreamRun() {
	atomic.StoreInt64(&lastStreamRunNs, int64(time.Since(processStart)))
}

func streamIsStale(interval int64) bool {
	if interval < 1 {
		return false
	}
	last := time.Duration(atomic.LoadInt64(&lastStreamRunNs))
	return time.Since(processStart)-last >= 2*time.Duration(interval)*time.Second
}

func handleStreamWatchdog(bouncer *Bouncer) {
	if bouncer.crowdsecMode != configuration.StreamMode && bouncer.crowdsecMode != configuration.AloneMode {
		return
	}
	if !streamIsStale(bouncer.updateInterval) {
		return
	}
	bouncer.log.Warn("handleStreamWatchdog:stale forcing a refresh")
	handleStreamTicker(bouncer)
}
```

with `markStreamRun()` at the top of `handleStreamTicker`. `time.Since(processStart)` keeps the monotonic reading on both ends, so clock steps don't matter, and there is nothing left that can latch.

## Smaller notes

- Keeping two trigger paths is right, but I'd argue for it the other way round from the description: the dedicated watchdog ticker is built exactly like the ticker that stalled, so if tickers are the problem it stalls too. **The metrics-ticker path is the one that earns its place**, because it's the one component we have evidence kept running during the incident.
- The new ticker is started inside the `if streamTicker == nil` guard at `bouncer.go:300`, which is an unsynchronised read of a package global written at `:312`; `go test -race` flags it under concurrent `New()`. With `streamStartupBlock` or alone mode there's a full LAPI round trip between the read and the write, which widens the window considerably. Not introduced here, but this PR adds a third ticker to that block and discards its stop channel. A `sync.Once` would settle it.
- `handleStreamWatchdog` runs the forced refresh synchronously inside `handleMetricsTicker`, delaying the metrics push by up to one LAPI round trip.
- `Test_handleStreamWatchdogRecoversStaleTicker` saves and restores four package globals and shares `cacheTimeoutKey` and the package-global ttl_map with `Test_handleStreamCache`. It passes, but it leaks state if it fails mid-way — and the latch leaking across tests is exactly what made my own repro tests interfere until I ran each in its own process.

Separately, and not this PR's job: `go test -race` on the base flags `isCrowdsecStreamStartup` (`bouncer.go:638`/`:654`) and `isCrowdsecStreamHealthy` (`:508`) under concurrent refreshes. The CAS currently makes them unreachable, which is why `-race` is clean here — but if the latch goes, they should become atomic.

I left some notes on the root cause over in #377 — short version, the "blocked stream call" hypothesis doesn't hold, so it's worth getting one more data point from the reporters before locking in the shape of the mitigation.

luigibarretta:
Thanks for the thorough review and, especially, for building the wedged-LAPI reproducer. I agree with the blocker: the CAS latch makes the current revision unsafe, so this version should not merge.

Your point about the existing `updated` cache lease is convincing. I will rework the branch around these invariants:

- remove `streamUpdateInProgress` entirely and keep the cache lease as the duplicate-request guard;
- use a monotonic heartbeat (`time.Since(processStart)`) rather than storing wall-clock Unix nanoseconds;
- add a regression test where one refresh never returns and prove that later ticks and the watchdog can still reach the LAPI;
- make the pre-existing stream health/startup flags race-safe when concurrent refreshes are re-enabled;
- address ticker initialization with `sync.Once` (including ownership of the stop channel), rather than extending the current unsynchronised global guard;
- keep the watchdog reachable from both the dedicated ticker and the metrics path, but avoid delaying the metrics report on the forced refresh;
- isolate package-global/cache state in the tests so a failing test cannot contaminate the next one.

I also agree with your reasoning about the two trigger paths: the metrics path is the independently observed liveness signal; the second ticker is only defense in depth and must not be presented as the stronger one.

I have added a fresh production observation to #377. Our deployment runs the plugin at `INFO`, so the requested `reportMetrics`, `alreadyUpdated`, and `New initialized` DEBUG records were not retained; I have been explicit about that limitation rather than treating absence as evidence.

I'll keep the external guarded restart watchdog in place during the rewrite and soak. Thanks again — this caught a real failure mode before the mitigation could create a worse one.

luigibarretta:
Pushed review fixes in 50cd7ae.

Changes:
- removed the process-wide in-progress latch; overlapping retries are governed by the existing cache lease, so a wedged refresh cannot permanently suppress recovery;
- replaced wall-clock timestamps with a monotonic process-relative heartbeat;
- made stream startup, health, and failure state atomic;
- protected stream/watchdog and metrics ticker initialization with sync.Once;
- made watchdog recovery asynchronous so metrics delivery is not delayed;
- made ticker ownership/stopping explicit and isolated global state in tests;
- added regressions for a permanently wedged first refresh, concurrent initialization, and metrics delivery during forced recovery.

Local verification:
- golangci-lint v1.63.4
- go test ./...
- go test -race ./...
- concurrency regressions under -race, count=10
- yaegi v0.16.1
- all 9 binary Traefik v3.7.11 + mock LAPI E2E scenarios

The branch is ready for another review. The fork has not been deployed to production.

luigibarretta:
Rebased the branch onto the current upstream `main` in `0137080` and resolved the only conflict in the test imports.

Fresh local validation on the rebased head:

- `go test ./...`
- `go test -race ./...`
- all 9 mock-LAPI E2E scenarios with the binary Traefik test harness

The four old workflow runs reported as failed did not execute jobs: they remained pending for fork-workflow approval and GitHub expired them after 30 days. The two runs on the new head currently show `action_required` for the same approval gate, rather than a test failure.

---

Posted by an automated maintainer delegated by @luigibarretta. If you would prefer Luigi to respond personally, please say so in the thread.

PR #390, opened 2026-09-05
⚡ batch the CIDR lookup into one read instead of one per prefix length

mathieuHa:
Targets `271-feature-support-decision-scope` rather than `main`, so it stacks on #368 rather than duplicating it.

Follows @mathieuHa's suggestion of batching the range lookup. Your `CIDRLookupKeys` + `cidrprefixlens` design already does the hard part — only probing prefix lengths that actually have a decision. This changes what it costs to probe them.

## The problem

```go
for _, key := range ip.CIDRLookupKeys(ipStr, parsePrefixLens(prefixLens)) {
    value, getErr := c.cache.get(cidrPrefix + key)
    if getErr == nil { return value, nil }
}
```

A hit short-circuits, but a **miss walks every candidate** — and a miss is the common case on the request path, since most requests are in no range at all. So the worst path is the hot path, and it grows every time a blocklist introduces a new prefix length.

## The change

The candidate keys are all known up front, so they go in one `MGet`. `CIDRLookupKeys` already returns them most-specific-first, so the first non-empty value is still the match; the result is unchanged.

| | reads per request |
| --- | --- |
| before | 1 + N |
| after | **2, independent of N** |

The remaining two are the `cidrprefixlens` read and the batched candidate read.

## Verified on a real redis

Against the redis in my cluster, counting server-side commands across 4 `GetCIDR` calls:

```
cmdstat_mget: calls +4        (exactly one per lookup, usec_per_call=22)
cmdstat_get:  +7              (the prefixlens read per lookup, plus 3 SetCIDR)
```

and the semantics hold, including most-specific-wins:

```
10.1.2.3     -> "c"    (the /16 captcha, not the /8 ban)
10.9.9.9     -> "t"
192.168.5.7  -> "t"
11.0.0.1     -> miss
```

## What is in it

- `mget` on `cacheInterface`, implemented with `MGet` on redis and by looping on the local cache, so the in-memory path is unchanged.
- `Test_GetCIDR_Reads` now asserts the batched cost, with a new case at **eight** prefix lengths to pin that it does not scale with N. That test previously asserted `wantReads: 4` for three prefix lengths — it is the thing being fixed, so updating it is the point rather than a workaround.

Green on `gofmt`, `go vet`, `go test ./...`, `golangci-lint` (pinned toolchain) and **`yaegi test`** on both the root and `pkg/cache` packages.

## Dependencies, and they are real

1. **maxlerebourg/simpleredis!8** for `MGet`. `go.mod` currently points at a pseudo-version of that branch (`v1.0.13-0.20260904085131-f8801cc098d2`) so this could be tested; **retarget to a release tag before merging.**
2. **#382**, the pointer-held readers. The pooled client holds a `sync.Mutex`, so `[]simpleredis.SimpleRedis` copies a lock. That change is carried here.

If you would rather not couple #368 to those, this can wait until both land — the `mget` hunk itself is about 25 lines.

## What I deliberately left out

@mathieuHa raised memoizing the prefix-length set in memory, which would take this to a single read. I would keep that separate: it needs the set refreshed from the stream tick rather than read per request, and it has to stay correct when a new prefix length appears mid-cycle. `SetCIDR` publishing the length before the decision is what makes that safe, but it is state worth adding on its own merits rather than smuggling in here.

Two other notes on the original idea, for the record:

- **`HGETALL` would be a regression.** Bucketed by prefix length, `cidr:24` could hold thousands of networks and you would transfer the whole bucket per request. `HGET cidr:24 <network>` is fine but costs the same N round trips as today.
- **Key patterns (`10.*`) would need `KEYS` or `SCAN`.** `KEYS` is O(keyspace) and blocks the server, which is dangerous at 24k keys; `SCAN` is cursor-based and multi-round-trip. Computing the exact candidate keys, which `CIDRLookupKeys` already does, is strictly better and `simpleredis` supports neither.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,5 +4,5 @@ go 1.22.12
 
 require (
 	github.com/leprosus/golang-ttl-map v1.1.7
-	github.com/maxlerebourg/simpleredis v1.0.12
+	github.com/maxlerebourg/simpleredis v1.0.13-0.20260904085131-f8801cc098d2
 )
--- a/go.sum
+++ b/go.sum
@@ -1,4 +1,4 @@
 github.com/leprosus/golang-ttl-map v1.1.7 h1:cF4AAFDDnJTFSV+/42sKLhmMluvLdRlCGS2UaifH6UM=
 github.com/leprosus/golang-ttl-map v1.1.7/go.mod h1:4QWHJPeVBbrkhOhXdhCv9IEiyj/YzkO04/iexy4vSe0=
-github.com/maxlerebourg/simpleredis v1.0.12 h1:VsJpk2l8U9QqxOWbYnEXbrs7iSNZFm2fNmFvlxYlQNk=
-github.com/maxlerebourg/simpleredis v1.0.12/go.mod h1:lT4LX02SOsE9PxUcSrz1QW5ZnO86gPbaiYBxmtcXEls=
+github.com/maxlerebourg/simpleredis v1.0.13-0.20260904085131-f8801cc098d2 h1:fSYQSdirObi0BVp+0BS8QTuY4SmHwpm588jz6bzsF20=
+github.com/maxlerebourg/simpleredis v1.0.13-0.20260904085131-f8801cc098d2/go.mod h1:lT4LX02SOsE9PxUcSrz1QW5ZnO86gPbaiYBxmtcXEls=
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -60,6 +60,17 @@ func (localCache) set(key, value string, duration int64) {
 	cache.Set(key, value, duration)
 }
 
+func (lc localCache) mget(keys []string) ([]string, error) {
+	values := make([]string, len(keys))
+	for i, key := range keys {
+		value, err := lc.get(key)
+		if err == nil {
+			values[i] = value
+		}
+	}
+	return values, nil
+}
+
 func (localCache) delete(key string) {
 	cache.Del(key)
 }
@@ -67,7 +78,7 @@ func (localCache) delete(key string) {
 type redisCache struct {
 	log     *slog.Logger
 	writer  simpleredis.SimpleRedis
-	readers []simpleredis.SimpleRedis
+	readers []*simpleredis.SimpleRedis
 	counter atomic.Uint64
 }
 
@@ -77,7 +88,7 @@ func (rc *redisCache) nextReader() *simpleredis.SimpleRedis {
 		return &rc.writer
 	}
 	idx := rc.counter.Add(1) % uint64(n)
-	return &rc.readers[idx]
+	return rc.readers[idx]
 }
 
 func (rc *redisCache) get(key string) (string, error) {
@@ -99,6 +110,27 @@ func (rc *redisCache) get(key string) (string, error) {
 	return "", errors.New(CacheMiss)
 }
 
+func (rc *redisCache) mget(keys []string) ([]string, error) {
+	raw, err := rc.nextReader().MGet(keys)
+	if err != nil {
+		switch err.Error() {
+		case simpleredis.RedisMiss:
+			return make([]string, len(keys)), nil
+		case simpleredis.RedisUnreachable:
+			return nil, errors.New(CacheUnreachable)
+		default:
+			return nil, err
+		}
+	}
+	values := make([]string, len(keys))
+	for i := range raw {
+		if i < len(values) {
+			values[i] = string(raw[i])
+		}
+	}
+	return values, nil
+}
+
 func (rc *redisCache) set(key, value string, duration int64) {
 	if err := rc.writer.Set(key, []byte(value), duration); err != nil {
 		rc.log.Error("cache:setDecisionRedisCache" + err.Error())
@@ -114,6 +146,7 @@ func (rc *redisCache) delete(key string) {
 type cacheInterface interface {
 	set(key, value string, duration int64)
 	get(key string) (string, error)
+	mget(keys []string) ([]string, error)
 	delete(key string)
 }
 
@@ -130,7 +163,7 @@ func (c *Client) New(log *slog.Logger, isRedis bool, writeHost string, readHosts
 		rc := &redisCache{log: log}
 		rc.writer.Init(writeHost, pass, database)
 		for _, h := range readHosts {
-			var r simpleredis.SimpleRedis
+			r := &simpleredis.SimpleRedis{}
 			r.Init(h, pass, database)
 			rc.readers = append(rc.readers, r)
 		}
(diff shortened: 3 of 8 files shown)


mathieuHa:
Ran this on a live cluster rather than only in tests. Traefik 3.7.10 with the branch loaded via `localPlugins`, ~50 rps of real HTTPS traffic through the bouncer, redis primary on the same node as traefik plus one replica a hop away, 60s per sample.

## Round trips are flat in N

Range decisions added through `cscli`, all outside the client subnet so every lookup is a **miss** — the common case and, before this change, the expensive one.

| prefix lengths | GET/req | MGET/req | MGET server time | traefik mean |
| --- | --- | --- | --- | --- |
| 0 | 2.00 | **0.00** | – | 12.70 ms |
| 1 (`/24`) | 2.00 | **1.00** | 2.8 µs | 16.70 ms |
| 3 (`/8 /16 /24`) | 2.00 | **1.00** | 3.5 µs | 16.58 ms |
| 7 (`/8 /12 /16 /20 /24 /28 /32`) | 2.00 | **1.00** | 6.7 µs | 19.87 ms |

Three round trips total regardless of N: the IP lookup, the `cidrprefixlens` read, and one batched `MGET`. On `271-feature-support-decision-scope` this is `2 + N`, so at seven prefix lengths that is **9 round trips instead of 3**.

The numbers hang together: the first `MGET` added ~4 ms of request latency, matching the ~5 ms per redis operation this deployment measures. Going from 1 to 7 keys added ~3 ms in total, where one round trip per length would have added roughly 30 ms.

## Matching still works

With a range covering one load generator pod's `/24`, holding everything else constant:

```
pod inside the banned /24   ->  20 403
pod outside it              ->  20 200
```

## Worth knowing: `cidrprefixlens` never shrinks

My first "zero ranges" sample came back with **1 MGET per request**. The cause was a leftover `cidrprefixlens` holding `8,16,24` from an earlier throwaway test, with a TTL of **315,329,106 seconds — ten years**.

`DeleteCIDR` drops the `cidr:` key but never removes the prefix length. The comment in #368 says that is deliberate, and I agree with the direction — dropping a length too early leaves decisions unmatched, which is the worse failure. Two consequences worth stating anyway:

- Once a prefix length is ever seen, every request probes it **forever**. There is no way to reclaim it short of deleting the key by hand, which I had to do twice while running this.
- **This change makes that trade much cheaper.** A stale length now costs one extra key inside an existing batch (microseconds) instead of one extra round trip (milliseconds). So it argues for this PR rather than against it — but operators with churn in their range sources should know the probe set only grows.

I would not change the behaviour here. Shrinking it safely needs the same care as memoizing the set, which I deliberately left out of this PR.

## Caveats

One deployment, one redis topology, and the absolute latencies are specific to it — the ~5 ms per operation is inter-node on Oracle ARM instances, so a same-node redis would show much smaller deltas. What should generalise is the shape: **round trips constant in N instead of linear**.

mathieuHa:
Some live numbers from running this branch in production for 3.8 days, in case they are useful.

**Setup.** Traefik 3.7 on a 7-node k3s cluster, plugin in `stream` mode with the redis cache
enabled and a read replica (`redisCacheReadHosts`), sustained ~50 rps of real traffic through
the middleware. Redis 8, primary and replica as separate pods.

**Round trips are flat.** Measured on the replica, which serves the reads:

```
cmdstat_get:calls=19348681,usec_per_call=3.60
cmdstat_mget:calls=9168319,usec_per_call=5.69
```

Live rate under the same load: `100.53 get/s` and `50.27 mget/s` against 50 rps, i.e. exactly
**2 GET + 1 MGET per request**. That ratio held regardless of how many distinct prefix lengths
were present in the decision set — which is the point of the change: the lookup went from
`1 + N` round trips to a constant 3.

**Cost per operation.** MGET is 5.69 µs against GET's 3.60 µs, so batching N prefix lookups
into one MGET costs ~58% more than a single GET while replacing N of them. Break-even is
under two prefix lengths; everything beyond that is a straight win.

**Write path dominates.** On the primary, `SET` is 57.85 µs/call versus 6.51 µs replaying on
the replica. Reads are cheap; the stream refresh is where the time goes.

**Read/write split.** 29.80M commands served by the replica against 1.58M on the primary, so
~95% of the load leaves the primary free for stream writes.

Sample is 9.17M MGETs and 19.35M GETs, so the ratio is not an artefact of a short run.

mathieuHa:
How the other CrowdSec bouncers match Range decisions, read from their source, for comparison with what this branch does.

| Bouncer | Decisions stored in | How a range is matched | Cost of a lookup that finds nothing |
| --- | --- | --- | --- |
| [lua-cs-bouncer](https://github.com/crowdsecurity/lua-cs-bouncer/blob/HEAD/lib/crowdsec.lua) (nginx, openresty) | nginx shared dict | IPs are stored as /32; every netmask of the family is probed, most specific first | 33 (IPv4) or 129 (IPv6) shared-dict reads |
| [caddy-crowdsec-bouncer](https://github.com/hslatman/caddy-crowdsec-bouncer/blob/HEAD/internal/core/store.go) | BART table in memory (`gaissmai/bart` via `hslatman/ipstore`) | longest-prefix match | no I/O |
| [cs-haproxy-spoa-bouncer](https://github.com/crowdsecurity/cs-haproxy-spoa-bouncer/tree/HEAD/pkg/dataset) | `sync.Map` for IPs, BART table behind an atomic pointer for ranges | exact IP first, then longest-prefix match | no I/O |
| [php-remediation-engine](https://github.com/crowdsecurity/php-remediation-engine/blob/HEAD/src/CacheStorage/AbstractCache.php) (redis, memcached, files) | shared cache | IPv4 only; each range is copied into every 256-address bucket it covers, so a /8 costs 65,536 writes | 1 bucket read, then a linear scan |
| [cs-cloudflare-worker-bouncer](https://github.com/crowdsecurity/cs-cloudflare-worker-bouncer/blob/HEAD/pkg/cloudflare/worker/worker.js) | Cloudflare KV | all ranges in one `IP_RANGES` JSON value, scanned linearly; the first match wins, not the most specific | 1 read, then a scan of every range |
| this branch | redis or local cache | candidate keys for the prefix lengths in use, most specific first, in one `MGET` | 1 `GET` + 1 `MGET` after the IP lookup, whatever the number of prefix lengths |

The Go bouncers keep everything in process memory, which does not fit a cache shared by several Traefik instances. The PHP engine is the only one with the same constraint, and it pays for its single read with write amplification and no IPv6 support.

The simplest design on paper is lua-cs-bouncer's: probe every length in one `MGET` and drop `cidrprefixlens` entirely. Measured under Yaegi on this branch, local cache, lookup that finds nothing:

| Lookup | µs per lookup |
| --- | --- |
| `GetCIDR` on this branch, IPv4, 7 prefix lengths | 136 |
| every IPv4 length (33 keys) | 294 |
| every IPv6 length (129 keys) | 1126 |

In the interpreter, building a key costs about 4 µs, which LuaJIT does not pay. The prefix-length set pays for itself here.

